PR #2, opened 2025-05-12
Disallow ccache

Szelethus:
This will fail anyways as bazel and ccache are famously incompatible or very hard to work together with.


dkrupp (approved):

nettle (commented):
To be honest I'm not quite sure we should do this check at least here and fail.
Ccache incompatibility is upper level Bazel problem.
Besides ccache might be disabled with CCACHE_DISABLE.

Szelethus:
> To be honest I'm not quite sure we should do this check at least here and fail. Ccache incompatibility is upper level Bazel problem. Besides ccache might be disabled with CCACHE_DISABLE.

As bazel wipes the environment immediately, `CCACHE_DISABLE` has no effect, but I could certainly try to create an exception for it so it can be forwarded. To be fair, this PR was more about helping to establish the proper environment for myself, but I also see this as a major headache for ccache users, as the error message is super cryptic.

How about I do that `CCACHE_DISABLE` forwarding and change the error message to draw attention to it?

src/clang.bzl
```diff
@@ -187,30 +187,32 @@ def _run_analyzer(
     )
     return outfile
 
+def check_valid_file_type(src):
```
nettle, line 190: This file should not be changed

src/clang_ctu.bzl
```diff
@@ -111,31 +111,32 @@ def _run_clang_ctu(
     )
     return outputs
 
+def check_valid_file_type(src):
```
nettle, line 114: This file should not be changed

src/code_checker.bzl
```diff
@@ -18,11 +18,21 @@ COMPILE_COMMANDS_JSON=$1
 shift
 COMPILE_COMMANDS_ABS=$COMPILE_COMMANDS_JSON.abs
 sed 's|"directory":"."|"directory":"'$(pwd)'"|g' $COMPILE_COMMANDS_JSON > $COMPILE_COMMANDS_ABS
-echo "Running: $@" $COMPILE_COMMANDS_ABS > $LOG_FILE
-echo "==================================" >> $LOG_FILE
+echo "CodeChecker command: $@" $COMPILE_COMMANDS_ABS > $LOG_FILE
```
nettle, line 21: This file should not be changed

src/code_checker.bzl
```diff
@@ -153,6 +153,9 @@ def _toolchain_flags(ctx, action_name = ACTION_NAMES.cpp_compile):
         feature_configuration = feature_configuration,
         action_name = action_name,
     )
+    if "ccache" in compiler:
```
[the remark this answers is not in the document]
nettle: If you still insist on checking for ccache, is it possible to add the check at repository rule level, just in one place (e.g. tools.bzl), not spreading it through all other rules?

nettle (changes requested):

PR #3, opened 2025-05-12
Code formatting and logs

Szelethus:


dkrupp (approved):
Nice. Logging is quite crucial.

src/code_checker.bzl
```diff
@@ -18,11 +18,21 @@ COMPILE_COMMANDS_JSON=$1
...
+echo "                   CodeChecker error log                   " >> $LOG_FILE
+echo "===-----------------------------------------------------===" >> $LOG_FILE
 eval "$@" $COMPILE_COMMANDS_ABS >> $LOG_FILE 2>&1
 # ls -la $DATA_DIR
 # NOTE: the following we do to get rid of md5 hash in plist file names
+ret_code=$?
+echo "===-----------------------------------------------------===" >> $LOG_FILE
+if [ $ret_code -eq 1 ]; then
+    echo "===-----------------------------------------------------==="
+    echo "[ERROR]: CodeChecker returned with -1!"
```
nettle, line 32: -1?

src/code_checker.bzl
```diff
@@ -18,11 +18,21 @@ COMPILE_COMMANDS_JSON=$1
...
+echo "CodeChecker command: $@" $COMPILE_COMMANDS_ABS > $LOG_FILE
+echo "===-----------------------------------------------------===" >> $LOG_FILE
+echo "                   CodeChecker error log                   " >> $LOG_FILE
+echo "===-----------------------------------------------------===" >> $LOG_FILE
 eval "$@" $COMPILE_COMMANDS_ABS >> $LOG_FILE 2>&1
 # ls -la $DATA_DIR
 # NOTE: the following we do to get rid of md5 hash in plist file names
+ret_code=$?
+echo "===-----------------------------------------------------===" >> $LOG_FILE
+if [ $ret_code -eq 1 ]; then
```
nettle, line 30: Just wonder if we should check for other non zero codes?

nettle (commented):
Just a small nitpick otherwise LGTM

src/code_checker.bzl
Szelethus: According to the CodeChecker docs (also, final few lines of `CodeChecker analyze --help`:
```
Exit status
------------------------------------------------
0 - Successful analysis
1 - CodeChecker error
3 - Analysis of at least one translation unit failed
128+signum - Terminating on a fatal signal whose number is signum
```
So it would make sense to check for >128.

PR #4, opened 2025-05-12
Pass the correct interpreter forward

Szelethus:
This caused a bug when bazel wiped the environment, and while the wrapper
script was invoked with the correct interpreter, when it called CodeChecker
through popen, it used the wrong interpreter from PATH. Lets just reuse the
interpreter we were invoked with.

Depends on #1, #3, #2.

### DIFF
--- a/src/codechecker_script.py
+++ b/src/codechecker_script.py
@@ -173,6 +173,8 @@ def analyze():
 
     output = execute("%s analyzers --details" % CODECHECKER_PATH, env=env)
     logging.debug("Analyzers:\n\n%s", output)
+    if "ccache" in output:
+        fail("ccache is not supported!")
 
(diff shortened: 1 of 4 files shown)


dkrupp (approved):

src/codechecker_script.py
```diff
@@ -126,6 +126,7 @@ def input_data():
 
 def execute(cmd, env=None, codes=[0]):
     """ Execute CodeChecker commands """
+    cmd = sys.executable + ' ' + cmd
```
nettle, line 129: This is actually might be quite dangerous!
I think we should check env requirements for CodeChecker: what is missing in Bazel sandbox?
BTW, there is also {PythonPath} which we may use.

PR #5, opened 2025-05-19
Add CodeChecker-like workflows

Szelethus:
We discussed that maybe the tests relating to these rules have a better
place here than in the CodeChecker repository itself. I mostly
copy-pasted the relevant files from the .github/ folder from CodeChecker
to this repo, and migrated the tests to pytest. Not sure if I aced any
logging info of value, that is worth the most attention in this PR.

Action link:
https://github.com/Szelethus/codechecker_bazel/actions/workflows/test.yml

As of now, depends on #1, #3.


.github/workflows/test.yml
```diff
@@ -0,0 +1,56 @@
...
+on: [push, pull_request]
+
+permissions: read-all
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # Note: UI related linter tests will run in the gui job.
```
nettle, line 13: Should be removed I guess

test/BUILD
```diff
@@ -149,6 +148,9 @@ codechecker_test(
     tags = [
         "manual",
     ],
+    analyze = [
+        "--analyzers clangsa clang-tidy",
```
nettle, line 152: Should not be here

test/test_rules.py
```diff
@@ -20,11 +20,10 @@
 import re
 import shlex
 import subprocess
-import sys
 import unittest
 
 
-class TestBase(unittest.TestCase):
+class TestBaseTest(unittest.TestCase):
```
nettle, line 26: Any reason for this change?

test/test_rules.py
```diff
@@ -191,23 +194,3 @@ def test_bazel_test_code_checker_ctu(self):
...
-
-def main():
-    """Run unittest"""
-    setup_logging()
-    logging.debug("Start testing...")
-    unittest.main(buffer=True)
-
-
-if __name__ == "__main__":
-    main()
```
nettle, line 213: Any reason to remove this very convenient possibility?

.github/workflows/install-deps.sh
```diff
@@ -0,0 +1,36 @@
...
+  clang-tidy-14 \
+  cppcheck
+
+pip install -U pytest
+
+# Source: https://fbinfer.com/docs/getting-started
+VERSION=1.1.0; \
+curl -sSL "https://github.com/facebook/infer/releases/download/v$VERSION/infer-linux64-v$VERSION.tar.xz" \
+| sudo tar -C /opt -xJ && \
+sudo ln -s "/opt/infer-linux64-v$VERSION/bin/infer" /usr/local/bin/infer
```
nettle, line 28: FB Infer, why do we need this?

.github/workflows/install-deps.sh
```diff
@@ -0,0 +1,36 @@
...
+  libc6-dev-i386 \
+  libpq-dev \
+  libldap2-dev \
+  libsasl2-dev \
+  libssl-dev \
+  clang-14 \
+  clang-tidy-14 \
+  cppcheck
+
+pip install -U pytest
```
nettle, line 22: Supporting pytest is good idea but should be introduced separately

.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -0,0 +1,32 @@
+---
```
nettle, line 1: I think this is not needed in this commit

.github/ISSUE_TEMPLATE/Feature_request.md
```diff
@@ -0,0 +1,19 @@
+---
```
nettle, line 1: I think this is not needed in this commit

test/BUILD
Szelethus: Are you sure? On machines where gcc static analyzer and infer are also enabled, or another one later down the line, the test fails, hence the limit.

.github/workflows/install-deps.sh
Szelethus: CodeChecker supports its, it'd be neat to test for it as well, but I guess it can wait.

.github/ISSUE_TEMPLATE/Bug_report.md
nettle: Well, at least separate PR. To be honest I dont think these files are needed at this point, maybe much later.

.github/ISSUE_TEMPLATE/Feature_request.md
nettle: At least separate PR. These files can be added much much later.

.github/workflows/install-deps.sh
```diff
@@ -0,0 +1,27 @@
...
+  clang-14 \
+  clang-tidy-14 \
+  cppcheck
+
+ldd --version
+
+sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-14 9999
+sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-14 9999
+sudo update-alternatives --install /usr/bin/clang-tidy clang-tidy /usr/bin/clang-tidy-14 9999
+sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-13 9999
```
nettle, line 27: These prerequisites looks way complex to me!
And they also look quite non-standard!
Do we really need all of these?
Are there explanation behind all of these?

.github/workflows/test.yml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: ubuntu-24.04
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
+
+      - name: Setup Bazel
+        uses: abhinavsingh/setup-bazel@v3
```
nettle, line 24: Is it official Bazel setup?

test/BUILD
nettle: We should support gcc in the rules in this case or disable it by default.

test/test_rules.py
```diff
@@ -24,7 +24,7 @@
 import unittest
 
 
-class TestBase(unittest.TestCase):
+class TestCodeCheckerBazelRules(unittest.TestCase):
```
nettle, line 27: Why do we need to rename this class?

test/test_rules.py
```diff
@@ -70,23 +70,14 @@ def grep_file(self, filename, regex):
...
+        with open(filename, "r", encoding="utf-8") as fileobj:
             for line in fileobj:
                 if pattern.search(line):
                     logging.debug(line)
                     return line
         self.fail(f"Could not find r'{regex}' in '{filename}'")
         return ""
 
-
-class TestBasic(TestBase):
```
nettle, line 82: The idea was to add more tests later

test/test_rules.py
```diff
@@ -118,25 +111,28 @@ def test_bazel_build_fail(self):
...
+            "compile_commands.json")
         self.grep_file(compile_commands, r"pass\.cc")
         self.grep_file(compile_commands, r"bin\/gcc")
 
     def test_bazel_aspect_clang_tidy_pass(self):
         """Test: bazel build :test_pass --aspects"""
-        command = "bazel build :test_pass"
-        command += " --aspects @bazel_codechecker//src:clang.bzl%clang_tidy_aspect"
-        command += " --output_groups=report"
+        command = "bazel build :test_pass " + \
```
nettle, line 124: Nitpick: to be honest I really dont like this kind of breaks :)

test/test_rules.py
```diff
@@ -157,18 +153,20 @@ def test_bazel_build_clang_analyze_fail(self):
...
         """Test: bazel test :clang_ctu_pass"""
         self.check_command("bazel test :clang_ctu_pass", exit_code=0)
 
     def test_bazel_test_clang_ctu_fail(self):
         """Test: bazel test :clang_ctu_fail"""
-        self.check_command("bazel test :clang_ctu_fail", exit_code=3)
-        logfile = os.path.join(
-            self.BAZEL_TESTLOGS_DIR, "clang_ctu_fail", "test.log")
-        self.grep_file(logfile, "// CTU example")
+        # FIXME: Currently failing in github CI.
```
nettle, line 165: Well, it is supposed to fail!
This comment means the test is not failing which is a sign

nettle (changes requested):
I suggest the change should be split to smaller, more granular and parts, so it will be easier to review and integrate. I would start from very simple GitOps integration with running very simple smoke tests.

test/BUILD
Szelethus: Very well. My thinking was that this is a test of the new bazel rules, not CodeChecker, so I figured its fine to have a test which is a little less fragile with different CodeChecker versions.

Szelethus: One more thing; we don't need to change the rules at all for each supported analyzer, correct?

.github/workflows/test.yml
Szelethus: The repo wiki page sure claims its doing the official install steps. In terms of the _official github workflow_ setup rules, this seems to be the closest, but it seems overkill and a real hassle to make it install the proper version of bazel:
https://github.com/bazel-contrib/setup-bazel

test/test_rules.py
Szelethus:
```
...........F.s....
======================================================================
FAIL: test_bazel_test_clang_ctu_fail (test_rules.TestBasic)
Test: bazel test :clang_ctu_fail
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/codechecker_bazel/codechecker_bazel/test/test_rules.py", line [...], in test_bazel_test_clang_ctu_fail
    self.grep_file(logfile, "// CTU example")
  File "/home/runner/work/codechecker_bazel/codechecker_bazel/test/test_rules.py", line 78, in grep_file
    self.fail(f"Could not find r'{regex}' in '{filename}'")
AssertionError: Could not find r'// CTU example' in '../bazel-testlogs/test/clang_ctu_fail/test.log'

----------------------------------------------------------------------
Ran [18](https://github.com/Szelethus/codechecker_bazel/actions/runs/15489740499/job/43612289356#step:7:19) tests in 29.775s
```

The error code is fine, the regex is the issue. This is specifically a github CI issue, it works fine locally.

Szelethus: I'll get this fixed. By the time I finished up the PR originally in the CodeChecker repo, this was the nth issue  I encountered which is why I decided to comment this out so I wouldn't need to stay downstream for so long.

Szelethus: Seems like the problem was, ironically, that `clang-extdef-mapping` was missing. I heard not even the CodeChecker tests have it set up for the CI, because its a bit of a hassle. My current understanding is that you can't install it from a package manager, so we only test CTU functionality in a different environment where the long compilation time of LLVM is more tolerable. I can see that on the official LLVM github repo, the large (!) tarballs have it in them, but the download+extract times are pretty bad.

PR #6, opened 2025-06-06
Add bug report template to github

Szelethus:


nettle (changes requested):
Can we drop this patch so far?
I dont think we will need this for quite time.

Szelethus:
I'm happy to drop the feature request thing, but I'd be happy to see the bug request form, especially since there are many components to get these rules going. Its important that the bazel version, CodeChecker version and this repos version is appropriately shown.

nettle (approved):

PR #8, opened 2025-06-17
Make line lengths <79 in test.py

Szelethus:
I also made sure to skip one of the tests if clang-extdef-mapping is missing, though maybe it should be promoted to a hard error in the rule definition.

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -19,7 +19,6 @@ import logging
 import os
 import re
 import shlex
-import shutil
 import subprocess
 import sys
 import unittest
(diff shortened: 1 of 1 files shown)


test/test.py
```diff
@@ -157,12 +162,15 @@ def test_bazel_build_clang_analyze_fail(self):
...
-        self.check_command("bazel build :clang_ctu_pass :clang_ctu_fail", exit_code=0)
+        self.check_command(
+            "bazel build :clang_ctu_pass :clang_ctu_fail", exit_code=0)
 
     def test_bazel_test_clang_ctu_pass(self):
         """Test: bazel test :clang_ctu_pass"""
         self.check_command("bazel test :clang_ctu_pass", exit_code=0)
 
+    @unittest.skipIf(shutil.which("clang-extdef-mapping") is None,
+                     "Missing 'clang-extdef-mapping")
```
nettle, line 173: Can you please remove this from the patch?
So we keep this problem separated.

nettle (changes requested):

PR #10, opened 2025-06-18
Install bazel 6.5.0 in CI through bazelisk

Szelethus:
Check results here:
https://github.com/Szelethus/codechecker_bazel/actions

PR #11, opened 2025-06-20
Install CodeChecker in CI

Szelethus:
See results here:
https://github.com/Szelethus/codechecker_bazel/actions

The eventual engoal of this PR series is to run the tests found in `test/test.py`. This has two notable dependencies: bazel (which was installed in #10, and CodeChecker, which this patch installs.

There are a variety of ways to install CodeChecker, the most notable is either compiling it from source or getting it from `python-pip`. The problem with the former solution is that the standard compilation method doesn't result in a self-contained, package-like CodeChecker. This would lead to the correct interpreter, on which the dependecies are installed in a virtual environment being lost along the way (see: #4).

As such, this patch installs CodeChecker through `python-pip`, where this is not an issue. Also, its a much simpler solution for now, we can always change to a top-of-the-tree CodeChecker built from source if we need to.

Note that `python-pip` doesn't install any of the analyzers (clangsa, clang-tidy, gcc, infer, cppcheck) it supports. A separate script installs all of these (as [discussed](https://github.com/Ericsson/codechecker_bazel/pull/5#discussion_r2150253230) in the previous PR, #5). I installed the latest versions for each available in Ubuntu 24.04's package manager, or, in the case of infer, got hold of the binary as per the official instructions.

Similarly to #10, the patch doesn't quite run the tests in `test/test.py`, it only prints out the versions of `bazel`, `CodeChecker`, and the version of the analyzers.


nettle (commented):
To be honest, again, this change looks too much complicated for me. I'm not sure I understand the goal.

Issue #7, opened 2025-06-17
Failed to run code_checker_test on flatbuffers


Szelethus:
## Describe the bug
Bazel build fails:
```bash
TERM=dumb bazel build //:code_checker_test_flatc --show_progress
```
Log (I manually reduced absolute paths):
```
Starting local Bazel server and connecting to it...
Loading: 
Loading: 
DEBUG: Rule 'bazel_codechecker' indicated that a canonical reproducible form can be obtained by modifying arguments commit = "289b9cefca68c6f23e8a87ca332a9231de9c24ab" and dropping ["branch"]
DEBUG: Repository bazel_codechecker instantiated at:
  flatbuffers/WORKSPACE:7:15: in <toplevel>
Repository rule git_repository defined at:

  /path/to/e0c365967510e2e55c8f330a5f2ef368/external/bazel_tools/tools/build_defs/repo/git.bzl:181:33: in <toplevel>
Loading: 
Loading: 
Loading: 
Loading: 
Loading: 
Loading: 
Loading: 
Loading: 0 packages loaded
Analyzing: target //:code_checker_test_flatc (1 packages loaded, 0 targets configured)
Analyzing: target //:code_checker_test_flatc (34 packages loaded, 25 targets configured)
Analyzing: target //:code_checker_test_flatc (34 packages loaded, 25 targets configured)
Analyzing: target //:code_checker_test_flatc (34 packages loaded, 25 targets configured)
Analyzing: target //:code_checker_test_flatc (34 packages loaded, 25 targets configured)
Analyzing: target //:code_checker_test_flatc (35 packages loaded, 25 targets configured)
Analyzing: target //:code_checker_test_flatc (42 packages loaded, 224 targets configured)
ERROR: flatbuffers/reflection/BUILD.bazel:1:10: in @bazel_codechecker//src:code_checker.bzl%compile_info_aspect aspect on filegroup rule //reflection:reflection_fbs_schema: 
Traceback (most recent call last):
	File "/path/to/e0c365967510e2e55c8f330a5f2ef368/external/bazel_codechecker/src/code_checker.bzl", line 214, column 33, in _compile_info_aspect_impl
		compilation_context = target[CcInfo].compilation_context
Error: <target //reflection:reflection_fbs_schema> (rule 'filegroup') doesn't contain declared provider 'CcInfo'
ERROR: Analysis of target '//reflection:reflection_fbs_schema' failed
[...]

PR #12, opened 2025-06-23
Fix issue with code_checker incorrectly running on header files

furtib:
Fix issue #7

I noticed when running CodeChecker through bazel that it tries to do analyses on header files, not only .cpp/.c files. Such analyses failed because there are no compilation commands associated with header files in compile_commands.json. The fix now skips headers.

This certainly fixes the problem immediately, but I'm not confident this is the right fix. Considering that we should analyze those files and only those we have a compilation command for, I wonder if we should instead just iterate over the compile_commands.json. Consider this PR as thinking out loud.

### DIFF
--- a/src/code_checker.bzl
+++ b/src/code_checker.bzl
@@ -103,7 +103,7 @@ def _run_code_checker(
 def check_valid_file_type(src):
     """
     Returns True if the file type matches one of the permitted
-    srcs file types for C and C++ header/source files.
+    srcs file types for C and C++ source files.
     """
     permitted_file_types = [
         ".c",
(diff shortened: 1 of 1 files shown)


src/code_checker.bzl
```diff
@@ -211,6 +211,8 @@ def _collect_all_sources(ctx):
     return sources
 
 def _compile_info_aspect_impl(target, ctx):
+    if not CcInfo in target:
+        return []
```
nettle, line 215: This looks quite correct

src/code_checker.bzl
```diff
@@ -310,6 +313,8 @@ def _code_checker_impl(ctx):
                 compilation_context = target[CcInfo].compilation_context
                 for src in srcs:
                     args = target[CompileInfo].arguments[src]
+                    if src.path.endswith(header_extensions):
+                        continue
```
nettle, line 317: However, this change is a bit questionable and needs to be checked: basically srcs should not contain header files, but if if does should we skip analysis for them?

furtib:
I have removed the header file extensions from check_valid_file_type since it is only used in  _collect_all_sources (through _rule_sources), which should only return source files anyway. This seems like a less hacky fix.

I have looked into codechecker.bzl, and it seems like we are using compile_commands.json there to get the source files, why don't we do that in code_checker.bzl too?

PR #16, opened 2025-07-07
Add a small test project to run codechecker_bazel on

furtib:
Created a ci job (small_test), to run the codechecker Bazel rule on a tiny Bazel project (yaml-cpp)

### DIFF
--- a/.github/workflows/patch-test-proj.sh
+++ b/.github/workflows/patch-test-proj.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+
+FORK_REPO="$1"
+FORK_BRANCH="$2"
+
+git clone --recurse https://github.com/jbeder/yaml-cpp.git
+cd yaml-cpp
+git checkout yaml-cpp-0.7.0
+
+# Specify bazel version for bazelisk
+echo "6.5.0" > .bazelversion
+
+# Add codechecker to the project
+cat <<EOF >> BUILD.bazel
+#-------------------------------------------------------
+
+# codechecker rules
+load(
+    "@bazel_codechecker//src:codechecker.bzl",
+    "codechecker_test",
+)
+load(
+    "@bazel_codechecker//src:code_checker.bzl",
+    "code_checker_test",
+)
+
+
+codechecker_test(
+    name = "codechecker_test_yaml-cpp",
+    targets = [
+        ":yaml-cpp",
+    ],
+)
+
+code_checker_test(
+    name = "code_checker_test_yaml-cpp",
+    targets = [
+        ":yaml-cpp",
+    ],
+)
+
+#-------------------------------------------------------
+EOF
+
+cat <<EOF >> WORKSPACE
+#----------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -48,3 +29,25 @@ jobs:
...
+        uses: ./.github/actions/env_setup
+
+      - name: Patch yaml-cpp
+        run:  |
+          sh .github/workflows/patch-test-proj.sh ${{ github.repository }} ${{ github.ref }}
+
+      - name: Run Bazel CodeChecker
+        run: |
+          cd yaml-cpp
+          bazel build :codechecker_test_yaml-cpp
```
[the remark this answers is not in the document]
furtib: Running it with bazel test produces errors (despite filling the log with the success message of CodeChecker), thats why I have been using bazel build.

.github/workflows/patch-test-proj.sh
```diff
@@ -0,0 +1,67 @@
...
+
+FORK_REPO="$1"
+FORK_BRANCH="$2"
+
+git clone --recurse https://github.com/jbeder/yaml-cpp.git
+cd yaml-cpp
+git checkout yaml-cpp-0.7.0
+
+# Specify bazel version for bazelisk
+echo "6.5.0" > .bazelversion
```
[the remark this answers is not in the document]
Szelethus: I'd consider coercing bazelisk to install bazel before this script runs -- either this .bazelversion file or the one genereted in the workflow is superfluos currently.

Another solution -- if this is to remain -- _copy_ the .bazelversion file that is generated in the workflow such that the version is not repeated.

.github/workflows/patch-test-proj.sh
Szelethus, line 1: Considering that we aim to add more than one project, we could call this file `patch-proj-yaml-cpp.sh"

.github/workflows/test.yml
```diff
@@ -48,3 +29,31 @@ jobs:
                "CodeChecker finds different analyzers when running in " \
                "bazel's sandbox environment!"
           CodeChecker analyzers
+
+  small_test:
```
Szelethus, line 33: Same here, lets call this yaml_cpp_test.

.github/workflows/test.yml
```diff
@@ -48,3 +29,31 @@ jobs:
...
+    # Test the bazel rules introduced by repository on an independent open-source project.
+    name: Test On Small Project
```
Szelethus, line 35: Test On Small Project: YAML Cpp

Szelethus (approved):
Please change "small project" to the name of the project consistently, as we will add more than one project. Otherwise LGTM!

nettle (changes requested):
Good approach! But I think this project (yaml-cpp) should not be the only one. I would suggest adding this project as one-of-many, like putting files into sub directories with patches and test launchers etc

furtib:
The composite action file's name must be action.yml; it can't be changed!

furtib:
This PR is getting too big, so I'm leaving it here as a template.

PR #24, opened 2025-07-24
Fix per-file incremental analysis

nettle:
Why: Incremental analysis doesn't work with the per-file rule

What: Make inputs dependent on headers only
Note: depending on headers only may mean that CTU wont work!

Test:
```
bazel clean --expunge
bazel build //test:code_checker_fail -s
vim test/src/pass.cc
bazel build //test:code_checker_fail -s
```

Addresses: #21

### DIFF
--- a/src/code_checker.bzl
+++ b/src/code_checker.bzl
@@ -320,7 +320,7 @@ def _code_checker_impl(ctx):
                         options,
                         compile_commands_json,
                         compilation_context,
-                        sources_and_headers
+                        sources_and_headers,
                     )
                     all_files += outputs
     ctx.actions.write(
--- a/src/code_checker.bzl
+++ b/src/code_checker.bzl
@@ -62,7 +62,7 @@ def _run_code_checker(
     clangsa_plist = ctx.actions.declare_file(clangsa_plist_file_name)
     codechecker_log = ctx.actions.declare_file(codechecker_log_file_name)
 
-    inputs = depset([compile_commands_json, src], transitive = [sources_and_headers])
+    inputs = depset([compile_commands_json, src], transitive = sources_and_headers)
     outputs = [clang_tidy_plist, clangsa_plist, codechecker_log]
 
     # Create CodeChecker wrapper script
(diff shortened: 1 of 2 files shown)


Szelethus:
Do I understand correctly that in this case if a header changes, still all TUs are reanalized, considering that each .cpp file is dependent on _all_ headers?

Or the idea is that the patch demonstrates that we are making bazel think of the analysis rules on a per-TU basis, where we can still refine the header set later down the line?

furtib (approved):

Szelethus (changes requested):
In terms of the high level stuff, this looks like a great step forward. Nice!

I see a few obstacles/future steps:
* We need some tests where we demonstate that changing a cpp file only triggers its own reanalysis, and one where changing a header (as of now) triggers a whole reanalysis
* I'd be happy to see the CI patches land (#37 at the very least) before this,
* We discussed in the related issue (#21) that in time, we should retrieve the true header dependency set. It sure sounds plausible (it really must be the case) that bazel knows the true header dependencies, even if the `hdrs` and `srcs` fields aren't the way to get them. Tibor mentioned some .d files that bazel generates but seem hard to access -- in loose, informal terms, how would you approach this problem? Make bazel list the header dependencies, should we have a compiler call, or ask the CodeChecker folks (which we will need to do for CTU anyways later on) to create a dry-run option?

nettle:
> * We need some tests where we demonstate that changing a cpp file only triggers its own reanalysis, and one where changing a header (as of now) triggers a whole reanalysis
> * I'd be happy to see the CI patches land (#37 or #28 at the very least) before this,

Totally agree with those.


> * We discussed in the related issue (#21) that in time, we should retrieve the true header dependency set. It sure sounds plausible (it really must be the case) that bazel knows the true header dependencies, even if the `hdrs` and `srcs` fields aren't the way to get them.

Well, these discussions are valid and we should continue experimenting with different approaches and ideas.
However, meanwhile I think we should submit this simple improvement even it is suboptimal but at least it is more Bazel-compatible.

> Tibor mentioned some .d files that bazel generates but seem hard to access -- in loose, informal terms, how would you approach this problem? Make bazel list the header dependencies, should we have a compiler call, or ask the CodeChecker folks (which we will need to do for CTU anyways later on) to create a dry-run option?

Bazel does not generate anything except artifacts described as Action output (clang `-MMD/-MD/-MM/-M` does).
So, we can describe an Action(s) to get whatever we need to know.
BTW, remember also that CodeChecker is just a wrapper around Clang.

Szelethus (approved):
LGTM! @nettle, I'd like to have your approval on this as well.

Issue #26, opened 2025-07-28
Test "test_bazel_test_fail" fails depending on installed analyzers!


furtib:
**Describe the bug**
In the test "test_bazel_test_fail", the number of warnings depends on the number of analyzers found on the testing system.
The test passes if only clang-tidy and clangsa are installed on the system.

**Versions**
Codechecker 6.26.0
Bazel 6.5.0
commit: e5fe33bb

We have a number of possible solutions to this:
We could:
- Make the check with grep, ignore the number of results
- Add the results of cppcheck to the checks
- **Restrict CodeChecker to only use clangsa and clang-tidy** (pass analyzers arg in BUILD)

This doesn't affect test "test_bazel_test_code_checker_fail" since it is run with the analyzers already restricted to clangsa and clang-tidy. (We should be careful since it will fail when we add new checkers in the future.)

furtib:
This is the only test that fails on my computer.

Szelethus:
I have some relevant discussion:

https://github.com/Ericsson/codechecker_bazel/pull/5#discussion_r2111980673

With respect to other analyzers, it requires some new changes (as discussed in https://github.com/Ericsson/codechecker_bazel/issues/17). I personally favour adding the tests first (possibly with some FIXMEs) and add analyzer support later. So I'd go with the third option.

nettle:
I'm sorry, what is this "test_bazel_test_codechecker_fail"?
How can I run this test? (at least run, since as I can understand it is not reproducible)

PR #13, opened 2025-06-24
Document all bazel rules provided in the repository

Szelethus:
I extended the README file to give a describe each rule currently found in the repository do and how can one load and use them. In particular, I wanted to draw attention to the difference in between codechecker_test and code_checker_test. Considering that many of these things are subject to change, I didn't wanna go overboard detailing each parameter yet.

This I hope is one PR among many aiming to improve our documentation. Considering that we are still in the early stages of this repository, the aim and scope are modest, and I'm not trying terribly hard to target a specific audience (soon-to-be-users, contributors, etc). Its simply an informal first impression page where every currently user-facing rules are described at least minimally.

There many other things that need to be done:
* Specify the minimal requirements and environment needed to pass all tests as they are now (https://github.com/Ericsson/codechecker_bazel/issues/29, #35 )
* A simple copy-paste command line to set up the repository and its requirements via apt, not only modules
* Have an actual user documentation, and update CONTRIBUTING.md.

Preview is here: https://github.com/Szelethus/codechecker_bazel/blob/readme_update/README.md

PR #28, opened 2025-07-28
CI for unit tests

furtib:
We should run the unit tests for every commit!

I have added the update-alternative command, because by default, `clang-extdef-mapping` was only accessible through `clang-extdef-mapping-18`

removed `--stat` option, opensource clang doesn't support this flag

While all tests are passing, keep #26 in mind! It only passes because we don't install cppcheck or Infer.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,3 +48,47 @@ jobs:
                "CodeChecker finds different analyzers when running in " \
                "bazel's sandbox environment!"
           CodeChecker analyzers
+
+  unit_test:
+    name: Unit Test
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set bazel version to 6.5.0
+        run: echo "6.5.0" > .bazelversion
+
+      - name: Setup Bazel
+        uses: bazel-contrib/setup-bazel@0.15.0
+
+      - name: Install python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Install CodeChecker analyzers
+        run: |
+          sudo apt-get update --quiet
+          sudo apt-get install --no-install-recommends \
+            clang \
+            clang-tools \
+            clang-tidy
+
+      - name: Install CodeChecker
+        run: pip3 install codechecker
+
+      - name: Print versions
+        run: |
+          bazel version
+          CodeChecker version
+          echo "[NOTE]: If you are debugging, its possible that " \
+               "CodeChecker finds different analyzers when running in " \
+               "bazel's sandbox environment!"
+          CodeChecker analyzers
+
+      - name: Run tests
+        run: |
+          cd test
+          python3 test.py -vvv
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,54 +36,16 @@ jobs:
             clang \
             clang-tools \
             clang-tidy
+          # clang-extdef-mapping is only available through
+          # clang-extdef-mapping-18 by default, but during ctu analysis
+          # we use it without version number (see clang_ctu.bzl).
+          # The sed command, extracts the version number from clang
+          # it reads the number from the word version to the first decimal point.
           sudo update-alternatives --install \
             /usr/bin/clang-extdef-mapping \
             clang-extdef-mapping \
-            /usr/bin/clang-extdef-mapping-$(clang --version | head -n 1 | sed -E 's/.*version ([0-9]+)\..*/\1/') \
-            100
-
-      - name: Install CodeChecker
-        run: pip3 install codechecker
-
-      - name: Print versions
-        run: |
-          bazel version
-          CodeChecker version
-          echo "[NOTE]: If you are debugging, its possible that " \
-               "CodeChecker finds different analyzers when running in " \
-               "bazel's sandbox environment!"
-          CodeChecker analyzers
-
-  unit_test:
-    name: Unit Test
-    runs-on: ubuntu-24.04
-
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4
-
-      - name: Set bazel version to 6.5.0
-        run: echo "6.5.0" > .bazelversion
-
-      - name: Setup Bazel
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -36,6 +36,11 @@ jobs:
             clang \
             clang-tools \
             clang-tidy
+          sudo update-alternatives --install \
+            /usr/bin/clang-extdef-mapping \
+            clang-extdef-mapping \
+            /usr/bin/clang-extdef-mapping-$(clang --version | head -n 1 | sed -E 's/.*version ([0-9]+)\..*/\1/') \
```
nettle, line 42: I guess, if there is no other way, we submit this workaround (I guess this is workaround, right?).
However, I think we should make sure:
1. We completely understand why we need this
2. We update documentation on how to do this locally (how other people fix the same problem)
3. We add detailed comment

.github/workflows/test.yml
```diff
@@ -48,3 +53,52 @@ jobs:
...
+      - name: Set bazel version to 6.5.0
+        run: echo "6.5.0" > .bazelversion
+
+      - name: Setup Bazel
+        uses: bazel-contrib/setup-bazel@0.15.0
+
+      - name: Install python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
```
nettle, line 74: Should we also update recommended prerequisites?

test/BUILD
```diff
@@ -134,7 +134,6 @@ codechecker_test(
     name = "codechecker_pass",
     analyze = [
         "--ctu",
-        "--stats",
```
nettle, line 137: Why? (There will be the second and the third "why" :) )

furtib: I removed the --stat option, because open-source clang doesn't support this flag.

.github/workflows/test.yml
furtib: We need this precisely because of this line:
https://github.com/Ericsson/codechecker_bazel/blob/e5fe33bbfa2ef7e4de18e3a5ca2325dbf368c790/src/clang_ctu.bzl#L288

.github/workflows/test.yml
Szelethus: This is not the earliest version we support, just didn't want to be [so old](https://github.com/Ericsson/codechecker_bazel/pull/5#discussion_r2150278269). We can submit another patch where we downgrade everything to the earliest possible, if you prefer (it will need many more `update-alternatives`  adjustments though).

.github/workflows/test.yml
```diff
@@ -48,3 +53,52 @@ jobs:
...
+
+  unit_test:
+    name: Unit Test
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set bazel version to 6.5.0
```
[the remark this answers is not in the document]
Szelethus: In this case, sure. I don't think there is a need or desire to support anything older than 6.5.0.

Szelethus (changes requested):
I like that the patch is leaner.

test/BUILD
nettle: As I can see it is CodeChecker option, is it not?
Besides, what open-source clang you mean? (we suppose to use Clang from CodeChecker)
(the second why :) )

Finally, if we want/need to remove this option, please create separate Issue (Steps to Reproduce) and separate Pull Request.

README.md
```diff
@@ -49,7 +49,7 @@ Prerequisites
 We need the following tools:
 
 - Git 2 or newer (we use 2.36)
-- Bazel 4 or newer (we recommend version 6)
+- Bazel 6 or newer (we recommend version 6)
```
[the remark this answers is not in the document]
nettle: Agree. But does it work with newer Bazel? 7? 8?

.github/workflows/test.yml
```diff
@@ -36,6 +36,18 @@ jobs:
             clang \
             clang-tools \
             clang-tidy
+          # The default naming of the clang-extdef-mapping, needed for CTU, is
+          # installed by clang-tools also contains the major version
+          # (e.g. clang-extdef-mapping-18), but the bazel rules reference it
+          # without the version number. To this end, we use update-alternatives
+          # to rename the binary to omit it.
```
nettle, line 43: So... you did this in your local setup? I didnt.
Should we explain this to the users in README and/or CONTRIBUTING?

nettle (changes requested):

test/BUILD
dkrupp: Only the ericsson downstream version of clang supports the statistics checker. CodeChecker only provides this option if it can detect that version of clang. So I think in this open source repo, we should not require it in the tests.

So I think it is okay to remove it here from the tests. Later, if we decide to open source the statistics checkers into clang we can reintroduce it.

nettle:
Please review #37

README.md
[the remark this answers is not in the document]
furtib: It does not work currently with Bazel 7.6.0 or 8.3.1
We can create support for them later down the line

.github/workflows/test.yml
furtib: Yes, I did do it on my local setup!
Ubuntu, by default, ships multiple versions of Clang and other related tools, and not all of the generic executable names are automatically linked to their versioned counterparts. The update-alternatives command is the standard way to handle this.

README.md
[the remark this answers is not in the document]
Szelethus: #13 will fix up the precise bazel version requirement.

.github/workflows/test.yml
Szelethus: This is actually the normal, if odd-looking solution on ubuntu. The desire to document is valid, though since this adjusment of binary name isn't the doing of this patch, I'd encourage tacking it separately as a part of #35.

README.md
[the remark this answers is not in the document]
Szelethus: Please remove this change, #13 fixed it and has already landed.

furtib:
I removed the line and also rebased from main

Szelethus (approved):

PR #37, opened 2025-07-30
Run tests in CI

nettle:
Why: we have GitHub action but it does not run any tests

What:
- Set Clang version (append particular llvm version to PATH)
- Check tools (which and versions)
- Run Bazel test
- Run unit tests
- Remove --stats (not supported by open-source CodeChecker)
- Change extension to yaml

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -18,33 +18,36 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - name: Set bazel version to 6.5.0
+      - name: Set clang version
+        run: echo "/usr/lib/llvm-18/bin" >> $GITHUB_PATH
+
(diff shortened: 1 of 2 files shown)


Szelethus (changes requested):
I'm personally strongly in favour of letting the original author finish the original work and learn in the process. It's fine to submit totally alterantive solutions to the same problem, we gained a lot from doing that in #21, but not quite the same solution with marginal differences.

`GITHUB_PATH` is interesting, but doesn't align with the, unfortunately only spoken, agreed on goals. The goal as I understand was to keep everything as simple as possible, representative of how a new user/contributor would set up their environment, which this GitHub Workflow specific solution doesn't achieve.

I think this should've been discussed as an idea in #28 and it is what I'd encourage moving forward.

Szelethus:
Landed in #28.

Issue #22, opened 2025-07-24
Define directory structure for tests


furtib:
We should discuss a better structure for our tests

nettle:
First draft to start discussion, just thinkin out loud: 
```
src/
test/
    foss/
        yaml-cpp/
            bazel.patch
            test.sh
        zlib/
            bazel.patch
            test.sh
    unit/
        avoidance/
        config/
        ctu/
        generated/
        platforms/
        simple/
            inc/
            src/
            BUILD
        transitive/
        virtual-includes/
            inc/
            src/
    test_unit.py
    test_foss.py
    test.sh
```
_FOSS = Free Open Source Software_

furtib:
I like the idea of moving the FOSS projects into the test directory!

I had an idea that we have a lot of repeating code in the initialization (formerly patch) scripts, so we could place them in a folder somewhere!

My idea for the FOSS part would be something like this:
`
foss/
├─ yaml-cpp/
│  ├─ init.sh
├─ zlib/
│  ├─ init.sh
├─ templates/
│  ├─ WORKSPACE.template
│  ├─ .bazelversion
`

What do you think?

Szelethus:
As we grow our testing infrastructure, I'd also be very much in favour of integrating [pytest](https://docs.pytest.org/en/stable/). To me, after switching to it in CodeChecker itself it emits a very easy-to-browse output.

nettle:
> As we grow our testing infrastructure, I'd also be very much in favour of integrating [pytest](https://docs.pytest.org/en/stable/). To me, after switching to it in CodeChecker itself it emits a very easy-to-browse output.

Pytest is good, but I would say only as an optional tool since it is one more external dependency.
I would like to keep compatibility with standard unittest

nettle:
> I had an idea that we have a lot of repeating code in the initialization (formerly patch) scripts, so we could place these repeating elements into a folder somewhere!
> ```
> foss/
> ├─ yaml-cpp/
> │  ├─ init.sh
> ├─ zlib/
> │  ├─ init.sh
> ├─ templates/
> │  ├─ WORKSPACE.template
> │  ├─ .bazelversion
> ```

Yes, `WORKSPACE.template` is definitely good idea! I guess we can also add `MODULE.template` later.
I'm not sure about `.bazelversion` because it might be easier to generate, but that's not that important.

BTW, why `init.sh`? What's your idea? And how do we run test(s)?

furtib:
[...]

PR #38, opened 2025-07-31
Restructure tests

furtib:
I created a prototype of the restructuring based on the current standing of #22 

I tried to group the tests logically.


furtib:
It's just the test_fail targets, aka: tests that find an error in an easy example. I didn't know what to call it, that's why I gave it this name.

Issue #41, opened 2025-08-04
Unit test doesn't pass if gcc is not in a bin folder


furtib:
**Describe the bug**
Our unit test, `test_bazel_compile_commands`, fails when the user uses a GCC that is not in a `bin` folder.
A use case for this would be if the user compiles GCC themselves and plans to use that.
This way gcc might be under a path like: `/home/$user/gcc-13/gcc`

**Versions**
* CodeChecker 6.26
* bazel: 6.5.0
* commit: e5fe33bbfa2ef7e4de18e3a5ca2325dbf368c790

**To Reproduce**
- Run this: `mkdir tmp; sudo ln -sf /usr/bin/gcc ./tmp/gcc && export PATH=$(pwd)/tmp/:$PATH `
- Run the unit tests

**Expected behaviour**
All tests pass

**Actual**
 `test_bazel_compile_commands` fails. Log:
```bash
python3 test/test.py
.........F...s....
======================================================================
FAIL: test_bazel_compile_commands (__main__.TestBasic)
Test: bazel build --build_tag_filters=compile_commands ...
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/test.py", line 128, in test_bazel_compile_commands
  File "test/test.py", line 78, in grep_file
AssertionError: Could not find r'bin\/gcc' in '../bazel-bin/test/compile_commands_pass/compile_commands.json'

----------------------------------------------------------------------
Ran 18 tests in 41.438s

FAILED (failures=1, skipped=1)
```

**Notes:**
Plan to fix:
- Remove the bin from the regex; this way GCC can be under any folder.

https://github.com/furtib/codechecker_bazel/blob/002ef0a12cbba56d2f8d956346a5d709f1f45c88/test/test.py#L128

Issue #36, opened 2025-07-29
Detect and disable ccache


nettle:
We need to find the best way to disable **ccache** (which must be disabled in Bazel)
especially for the environments where it is not possible to remove (`sudo apt-get remove`).

Scope
---

1. Experiment and suggest ways to disable ccache
   - Note that CCACHE_DISABLE=1 does not work
2. Describe the best way to disable ccache
3. Detect ccache in the code (tools.bzl?)
4. Update documentation and code
   - Add description to ccache detection code (tools.bzl?)
   - README
   - CONTRIBUTING (if needed)

DoD (Definition of Done)
---
* Detection is added
* Documentation is updated

nettle:
> I am currently unable to reproduce this issue! I am on Ubuntu 24.04, I have ccache installed, and following the [documentation](https://ccache.dev/manual/latest.html), I even "let ccache masquerade as the compiler" (I have done this to GCC, g++, clang and clang++)

Hm... that's strange.
Can you share the following:
```
which gcc
which g++
which clang
```

furtib:
After a reboot, now I do have this error!!
Checking with our unit tests, it seems like it only affects the `clang_*` rules

furtib:
Just thinking out loud:
Adding the flag `--action_env=CCACHE_DISABLE=1` fixes the problem.
Adding `build --action_env=CCACHE_DISABLE=1` to the .bazelrc file of the project fixes the problem. (it adds this flag to all build jobs)
Can we manipulate the environment with our rules directly?

nettle:
> Ofc!
> 
> [efurtib|~]$ which gcc
> /usr/local/bin/gcc
> [efurtib|~]$ which g++
> /usr/local/bin/g++
> [efurtib|~]$ which clang
> /usr/local/bin/clang
> [efurtib|~]$ which clang++
> /usr/local/bin/clang++
> The original binaries are under `/usr/bin`

This is quite unusual :) Maybe something new in ccache?
Typical (or old?) path to ccache is `/usr/lib64/ccache/` or similar (at least I used see such kind of configuration in Ubuntu).
Can you also check `ls -la /usr/local/bin/gcc` please?

In RHEL I have the following:

```
which gcc
/usr/lib64/ccache/gcc

which g++
/usr/lib64/ccache/g++

which clang
/app/vbuild/RHEL8-x86_64/clang/16.0.3/bin/clang
# I use clang from module load
```

[...]

PR #42, opened 2025-08-04
Accept gcc from any folder not just `bin`

furtib:
Why: Test `test_bazel_compile_commands`  fails when gcc is not in a `bin` folder.
What: Change the regex to accept gcc from any folder

Fixes issue: #41


nettle:
> What do you mean we will miss ccache? Doesn't ccache use a link with the same name as the compiler (in this case, GCC)? Please elaborate!

I mean in some (at least seen in RHEL) environments compiler may look like /usr/lib64/ccache/gcc
Please see #36

furtib:
Oh, I see.

However, this patch catches that case, well, if we look past the fact that it wouldn't run at all with ccache.
I updated the regex from `bin/gcc` to just `/gcc`.
So it would match `/usr/lib64/ccache"/gcc"`

nettle:
> So it would match `/usr/lib64/ccache"/gcc"`

I guess I was not clear enough.
I mean currently, in case of ccache usage which is logged as `/usr/lib64/ccache/gcc` the test would fail, right?
And it should fail!
If we remove `bin` from regex the test wont fail in case of ccache.

BTW, Sorry, I actually missed the point, what do we fix in this patch?
Why do we need to accept gcc from any folder not just bin? Do we have use-cases in #41?
Sorry, I just forgot.

Szelethus (commented):
I understand the issue was edited as well to be a bit more precise.

Based on what I explained above, I am in favour of accepting this patch. @furtib, any ideas how we could potentially test this? I know these things are notoriously difficult, so in this instance I wouldn't insist.

Szelethus (approved):
Meant to be an approve oops.

Issue #50, opened 2025-08-08
Run unittests on diferent ditributions


furtib:
## Describe the task
We should run our unittests on different distributions (mainly RHEL 9 and 8).
We should agree on which RHEL versions we should test on!

## Scope
- Add a CI job that runs unit tests on a RHEL image

## Definition of Done
Agreeing on which RHEL versions we should test on. 
Having a unit test job for RHEL machines

PR #51, opened 2025-08-08
Add unittest on rhel9

furtib:
## Why
We want to support RHEL systems, so we should run our unit tests on RHEL systems too.

## What
Added a new CI job that sets up the environment for a RHEL system and runs unittest on it.

## Scope

## Addresses
#50

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -106,9 +106,9 @@ jobs:
         run: |
           bazel version
           CodeChecker version
-          echo "[NOTE]: If you are debugging, its possible that " \
-               "CodeChecker finds different analyzers when running in " \
-               "bazel's sandbox environment!"
+          echo "[NOTE]: CodeChecker may find different analyzer binaries" \
+               "when invoking directly, or in bazel's sandbox environment!" \
+               "Be sure to double check during debugging."
           CodeChecker analyzers
 
       - name: Run tests


.github/workflows/test.yml
```diff
@@ -65,3 +64,53 @@ jobs:
...
+        run: pip3 install codechecker
+
+      - name: Print versions
+        run: |
+          bazel version
+          CodeChecker version
+          echo "[NOTE]: If you are debugging, its possible that " \
+               "CodeChecker finds different analyzers when running in " \
+               "bazel's sandbox environment!"
+          CodeChecker analyzers
```
nettle, line 111: Nitpick: why do we need this then?

.github/workflows/test.yml
```diff
@@ -65,3 +64,53 @@ jobs:
         run: |
           cd test
           python3 test.py -vvv
+
+  rhel9_test:
```
nettle, line 68: Warning: file keeps growing and will not be readable soon

nettle (commented):
Good!

.github/workflows/test.yml
```diff
@@ -65,3 +64,53 @@ jobs:
...
+    name: Unit tests on RHEL9
+    runs-on: redhat/ubi9
```
nettle, line 70: But... Are you sure it is a valid platform?

.github/workflows/test.yml
Szelethus: Considering that this, in my experience, one of the more frustrating parts of debugging, I'd be happy if we left in some warning, even if its not pretty. If it even once saves 3 hours of debugging for someone, its worth it. I will concede this isn't the perfect warning message, so lets massage it a bit and keep it.

Szelethus (commented):
Lets fix up that warning message a bit. I agree with the file length concern, but I'm also fine landing a fix later on.

furtib:
I have started a discussion in #55 about the length concern. I think leaving the unit tests together should be fine, and if we decide to split them up, we can always do it later. Regarding the Note, what if we fix the path by setting it in `package_layout.json`? this way we can be sure about the CodeChecker analyzers PATH.

.github/workflows/unit_test.yml
```diff
@@ -47,11 +47,61 @@ jobs:
...
       - name: Install CodeChecker
         run: pip3 install codechecker
 
+      - name: Print versions
+        run: |
+          bazel version
+          CodeChecker version
+          echo "[NOTE]: CodeChecker may find different analyzer binaries" \
+               "when invoking directly, or in bazel's sandbox environment!" \
+               "Be sure to double check during debugging."
```
Szelethus, line 60: Please copy-paste this over the old warning in the rhel9 job.

Szelethus (approved):
LGTM with a nit. I am in favour of keeping the warning, I don't find it as distracting as much I foresee it saving out behinds in one instance or another.

Szelethus:
@nettle I'd be happy to have your approval as well. I absolutely agree with the warning that we should keep the length (and complexity) of the file in check, but I think it'd be valuable to let this PR through so we easily check RHEL9 applicability on other patches until we hash this out.

Issue #35, opened 2025-07-29
Add Getting Started


nettle:
We should add quick **Getting Started** section (probably instead of Prerequisites, or make Prerequisites as a part of Getting Started) to make sure that environment can be set correctly.

Scope
---

1. List supported environments (host platforms), e.g.:
   - Ubuntu (versions?)
   - RHEL (versions?)
2. Reveal all environment requirements and define prerequisites (#29)
   - CodeChecker, Bazel, Python, Clang (versions?)
   - Disable/remove ccache?
   - Rename clang-extdef-mapping?
   - What else?
3. Develop simple shell commands for supported environments
4. Test environment setup
5. Update documentation
   - README
   - CONTRIBUTING (if needed)

DoD (Definition of Done)
---
Documentation is updated and tested

PR #52, opened 2025-08-08
Add getting started section

furtib:
## Why
We need a guide for new users to set up their environments.

## What
I added environment setup steps to README.

## Scope

## Addresses
#35

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,12 +74,12 @@ Install dependencies:
 
 RHEL 9:
 ```bash
-dnf update -y && dnf install -y llvm-toolset clang-tools-extra git python3 python3-pip gcc g++
+dnf update -y && dnf install -y wget llvm-toolset clang-tools-extra git python3 python3-pip gcc g++
 ```
 
 Ubuntu:
 ```bash
-sudo apt-get update --quiet && sudo apt-get install --no-install-recommends git python3 python3-pip python3-venv gcc g++ clang clang-tools clang-tidy
+sudo apt-get update --quiet && sudo apt-get install --no-install-recommends wget git python3 python3-pip python3-venv gcc g++ clang clang-tools clang-tidy
 ```
 
 On some distributions, `clang`, `clang-tidy` and `clang-extdef-mapping` may be installed with a trailing version number (e.g. clang-extdef-mapping-18). In case your package didn't install a non-versioned symlink as well, you will need to manually change it:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,6 +67,41 @@ are available in your system, you can just add the following modules:
...
+```bash
+update-alternatives --install /usr/bin/clang-extdef-mapping clang-extdef-mapping /usr/bin/clang-extdef-mapping-18 100
+update-alternatives --install /usr/bin/clang-tidy clang-tidy /usr/bin/clang-tidy-18 100
+```
+
+Install CodeChecker:
+```bash
+python3 -m venv ./codechecker_venv
+source ./codechecker_venv
+pip3 install codechecker
```
furtib, line 99: Is there a better way to install CodeChecker?

README.md
```diff
@@ -67,6 +67,41 @@ are available in your system, you can just add the following modules:
...
+```bash
+dnf update -y && dnf install -y llvm-toolset clang-tools-extra git python3 python3-pip gcc g++
+```
+
+Ubuntu:
+```bash
+sudo apt-get update --quiet && sudo apt-get install --no-install-recommends git python3 python3-pip python3-venv gcc g++ clang clang-tools clang-tidy
+```
+
+On ubuntu you have to make sure that `clang`, `clang-tidy` and `clang-ext-defmapping` commands are available! If they have version number appended to them use update-alternatives to get rid of the version number!
```
Szelethus, line 89: You use a little too many exclamation points. This is the way I like to write as well, put some passion in an otherwise dry communication medium, but this is not the place.

Try to keep your documentation's tone a bit more formal. Reserve exclamation points for truly exceptional places (warnings, new feature announcements), that way they also carry more meaning.

README.md
```diff
@@ -67,6 +67,41 @@ are available in your system, you can just add the following modules:
     module add codechecker/6.23
 
 
+Getting Started
+---------------
+
+Supported environments:
+- RHEL 9
+- Ubuntu 24.04 LTS
```
Szelethus, line 75: I mean, we don't _not_ support other platforms -- its just these are the ones we are dedicating the bulk of our attention to. How about "our ruleset was tested on the following operating systems, but you are welcome to try others"? Also, I work on Ubuntu 22, and it works just fine.

README.md
Szelethus: How about this:
"Some distibutions of clang, clang-tidy and clang-extdef-mapping may be installed with a trailing version number (e.g. clang-18). In case your package didn't install a non-versioned symlink as well, you will need to manually change it:"

README.md
Szelethus: Why is there a need for venv? Is this best practice?

Also, this is likely not the topic of a quick getting started section, but we should mention that this is practically the _only_ way to get CodeChecker working, as compiling from source introduces an environmental catastrophe  and everything will break.

README.md
```diff
@@ -67,6 +67,41 @@ are available in your system, you can just add the following modules:
...
+
+Install CodeChecker:
+```bash
+python3 -m venv ./codechecker_venv
+source ./codechecker_venv
+pip3 install codechecker
+```
+
+### WARNING! Remove ccache!
+The rules don't work with ccache present on the system!
```
Szelethus, line 103: Prefer using alerts:
https://github.com/lifeparticle/Markdown-Cheatsheet?tab=readme-ov-file#alerts-8-january-2024

Szelethus (changes requested):
Do you not need to install bazel?

README.md
furtib: You cannot install packages with pip system wide (or at least I'm not aware). This is considered best practice, regarding python packages.

I will create a warning for it!

README.md
furtib: I don't think adding ubuntu 22 is a good idea, maybe if we added a ubuntu 22 machine to the CI...

PR #48, opened 2025-08-07
Create a task issue template

Szelethus:
## Why

People who submit tasks need to follow some structure so that it's clear what the task is about, what are the action points in it, and when an issue can be regarded as completed.

## What

Add a general description, scope and DoD fields to the github template.

## Scope

The scope is minimal and a single accept should suffice.

## Addresses

The task template component of #34.

Issue #34, opened 2025-07-29
Define Issue process and requirements


nettle:
We should define and agree on Issue process and requirements to decrease the entropy level when working with Issues :)

Scope
---

1. Agree on Issue types, e.g.:
   - Bug, Feature, Task
2. Define Mandatory and Recommended items for each Issue type, e.g.:
   - Bug: Precondition, Steps to Reproduce, Expected, Actual
   - Feature: User-case description (what else?)
   - Task: Summary, Scope, DoD (Definition of Done)
3. Discuss and agree
4. Update documentation
   - CONTRIBUTING
   - Issue templates
   - README (if required)

DoD (Definition of Done)
---
* Process and Requirements are discussed and agreed
* Relevant documentation is updated

Notes:
---
I guess we might re-use CodeChecker process.

nettle:
Yes, looks fine to me :) And, I would say this process is needed mostly for you.
The main idea is good: issue -> discuss -> PR

I will also add a few suggestions to the bug template, probably to feature request and task as well.

PR #53, opened 2025-08-11
Create a task issue template

Szelethus:
Why: People who submit tasks need to follow some structure so that it's clear what the task is about, what are the action points in it, and when an issue can be regarded as completed.

What: 
* Add a general description, scope and DoD fields to the github template.
* Its quite important to do this under the original repo (under the `Ericsson` org), because issue types are an organizational setting, and if done under my own fork (under `Szelethus`) [I'm getting a warning](https://github.com/Szelethus/codechecker_bazel/blob/task_issue_template/.github/ISSUE_TEMPLATE/Task.yaml).
* For a peek, see the file here: https://github.com/Ericsson/codechecker_bazel/blob/issue-template-test/.github/ISSUE_TEMPLATE/Task.yaml
* Supersedes #48 because the source branch was changed.

Scope: The scope is minimal and a single accept should suffice.

Addresses: The task template component of #34.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Task.yaml
+++ b/.github/ISSUE_TEMPLATE/Task.yaml
@@ -1,0 +1,27 @@
+name: Task
+description: Create a task that needs to be discussed, investigated, or implemented.
+type: task
+body:
+  - type: textarea
+    id: describe-task
+    attributes:
+      label: Describe the task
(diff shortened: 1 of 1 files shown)


furtib (approved):
LGTM!

PR #60, opened 2025-08-13
Move old unit test into a legacy folder

furtib:
Why: We want to reorder our tests, and as a first step, we want to move the old tests to a legacy folder, so we can go forward with adding new tests to the suite.

What:
- Moved the old test files into unit/legacy
- Modified the BAZEL_BIN_DIR and BAZEL_TESTLOGS_DIR to account for these changes
- Created a wrapper script that finds and runs all tests found in the unit directory

Scope:

Addresses:
#22 

Depends on: #62


Szelethus (commented):
Lets think on this a little. It sounds too bad to be true that we need a script to run all tests in an idiomatically set up directory structure. Something like 
```
python -m unittest discover test/ -vvv
```
but recursive isn't an option? I know that pytest can handle this no problem.

test/unit/legacy/legacy_test.py
```diff
@@ -0,0 +1,221 @@
...
+import unittest
+
+
+class TestBase(unittest.TestCase):
+    """Unittest base abstract class"""
+
+    BAZEL_BIN_DIR = os.path.join("..", "..", "..", "bazel-bin", "test", "unit",
+                                 "legacy")
+    BAZEL_TESTLOGS_DIR = os.path.join("..", "..", "..", "bazel-testlogs",
+                                      "test", "unit", "legacy")
```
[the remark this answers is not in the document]
furtib: yes

furtib:
Your command didn't work because it expected file names to follow `test_*.py`.

test/test.sh
```diff
@@ -0,0 +1 @@
+python3 -m unittest discover unit $@
```
Szelethus, line 1: Lets add a [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)#Examples).

Szelethus (approved):
LGTM with a nit.

.github/workflows/unit_test.yaml
```diff
@@ -73,4 +73,4 @@ jobs:
       - name: Run tests
         run: |
           cd test
-          python3 test.py -vvv
+          python3 -m unittest discover unit -vvv
```
Szelethus, line 76: The output doesn't respect the `-vvv` flag: https://github.com/Ericsson/codechecker_bazel/actions/runs/17065925489/job/48382962683?pr=60

Szelethus (changes requested):

Szelethus (approved):
Lets send it.

Issue #55, opened 2025-08-11
Discussion: CI structure


furtib:
We should agree on a structure for the layout of the CI jobs in the .github folder!

Currently, all of our CI jobs reside in `.github/workflows/test.yml`. In the long term, this will result in an extremely long file, which is impractical to maintain.

# Scope
1. Agree on a directory/file structure for our CI files!

# DoD
1. We agree on a file structure
2. A PR implementing these changes is merged

furtib:
After a bit more thought, It might make sense to organize them in a way like this:
```
.
└── .github/
    └── workflows/
        ├── unit/
        │   ├── ubuntu.yml
        │   └── rhel9.yml
        └── foss/
            ├── ubuntu.yml
            └── rhel9.yml
```

furtib:
This could work, but the setup part should be a composite action (this is the requirement for it to be reusable).
Composite actions have a set of restrictions on them, so it would look more like this:
(Restrictions, such as: yml file name must be action.yml)
```
.
└── .github/
    ├── actions/
    │   ├── ubuntu_setup/
    │   │   └── action.yml
    │   └── rhel9_setup/
    │       └── action.yml
    └── workflows/
        ├── unit.yml
        └── foss.yml
```
We don't have to place the folders ubuntu_setup and rhel9_setup into the actions folder; it's just usually done this way.

Also, I'm in favour of using a Docker image instead of setting up the environment by hand each time.

PR #62, opened 2025-08-15
CI refactor

furtib:
Why:
We should try to minimize the size of CI job files

What:
I have created a composite action file for each OS we are testing on, so it can be reused in unit_test.yml and foss.yml.
I have modified the unit_test.yml file to use these composite actions.

Addresses:
#55 

Depends on:
#51


nettle:
Two minor questions, if I may:
1. It looks quite clumsy to have naming like **actions/something/action.yml**
   Is it possible to have one (or better zero!) "action" in the path?
2. Can we use original (not DOS) file extensions, e.g. **yaml** instead of yml?

furtib:
Rebase solved the problem!

PR #63, opened 2025-08-19
Warn on ccache

### LINKED ISSUE #36: Detect and disable ccache

furtib:
Why:
Bazel and ccache don't work well together. We should warn the user that their analyzers use ccache and fail the job to force the user to disable ccache (by changing package_layout.json or uninstalling ccache)

What:
Added a new job that checks if any checkers found by CodeChecker are symlinks to ccache. If any are, the job won't produce its output file, failing the build/test.

Note:
Should we, or shouldn't we give the output of the ccache checking job to all other jobs? If we don't give it as input, it might not run before the other jobs.

Addresses:
Fixes #36

### DIFF
--- a/src/code_checker.bzl
+++ b/src/code_checker.bzl
@@ -297,7 +297,29 @@ def _code_checker_impl(ctx):
         inputs = [],
         outputs = [ccache],
         mnemonic = "CCACHE",
-        command = "if ccache --version > /dev/null; then echo ERROR CCACHE FOUND; else echo no-ccache > ccache.txt; fi",
+        arguments = [ccache.path],
+        command = """
+        RED="\\033[0;31m"
+        BOLD="\\033[1m"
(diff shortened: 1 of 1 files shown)


furtib:
After a discussion, we decided that this is not the way forward.

Issue #64, opened 2025-08-25
Running on different sandbox levels


furtib:
**Describe the bug**
Resolution of softlinks in plist files in the codechecker rule fails, depending on the sandbox level used by Bazel.
With certain files (header files, but there could be examples with source files too), the path left in the list file is completely wrong; it looks something like this: `bazel_codechecker/bazel-out/k8-fastbuild/...`

**Versions**
* `CodeChecker 6.26`
* `bazel 6.5.0`
* on branch: [furtib:fix_plist_paths](https://github.com/furtib/codechecker_bazel/tree/fix_plist_paths) commit: 19600a9ada385e073c3ff32df9af0d11a0dafd7d

**To Reproduce**
To reproduce, try:
```bash
bazel build //test/unit/virtual_include:codechecker_virtual_include  --spawn_strategy=processwrapper-sandbox
```
Then, examine the generated list files.

Script to set up from scratch:
```bash
git clone https://github.com/furtib/codechecker_bazel.git
cd codechecker_bazel
git checkout 19600a9ada385e073c3ff32df9af0d11a0dafd7d
USE_BAZEL_VERSION=6.5.0 bazel build //test/unit/virtual_include:codechecker_virtual_include  --spawn_strategy=processwrapper-sandbox
```

**Expected behaviour**
I expected the path to be resolved to the original file

**Additional context**
Docker images use processwrapper-sandbox while local machines use linux-sandbox by default

PR #67, opened 2025-08-26
Add warning about Docker in getting started section

furtib:
Why:
I had a very hard time figuring out why things worked differently in Docker. Now that I have found the answer to this question, I want to preserve it.

What:
Added a warning about Docker using different sandbox environments.

Addresses:
#64

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ update-alternatives --install /usr/bin/clang clang /usr/bin/clang-18 100
 
 Install CodeChecker:
 > [!Note]
-> Currently the pip installed CodeChecker works best with these rules, which is our recommendation (as opposed to using acustom-built CodeChecker).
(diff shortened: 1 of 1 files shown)


Szelethus (commented):
I agree with the warning, not sure about the placement. We're still a bit weak on separating what belongs to README, CONTRIBUTING, etc. This feels like it'd have a better place in the latter.

Szelethus (approved):
Considering the exploratory state of these rules for now, I think this is a reasonable point of caution for folks testing or future contributors. No, this is absolutely not the perfect place for this warning, but we should reorganize these docs anyways.

My view is that it is better to have this warning somewhere, than still being stuck in a stale PR. Lets remember to speed up our docs refactoring tasks.

PR #71, opened 2025-09-01
Add ccache detection

### LINKED ISSUE #36: Detect and disable ccache

furtib:
Why:
Bazel and ccache don't work well together. We should warn the user that their analyzers use ccache and fail the job to force the user to disable ccache (by changing package_layout.json or uninstalling ccache)

What:
Extended the repository rule that detects CodeChecker to also detect ccache

Addresses:
Fixes https://github.com/Ericsson/codechecker_bazel/issues/36

### DIFF
--- a/src/clang.bzl
+++ b/src/clang.bzl
@@ -1,5 +1,9 @@
 load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "ACTION_NAMES")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
+load(
+    "@default_codechecker_tools//:defs.bzl",
+    "CCACHE_DISABLE"
+)
 
 CLANG_TIDY_WRAPPER_SCRIPT = """#!/usr/bin/env bash
 CLANG_TIDY=$1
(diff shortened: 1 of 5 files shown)


Szelethus (commented):
I want to give some credit, because this really is a novel and clever solution. I just happen not to prefer it to `--ccache-skip`, because we can't guarantee that ccache will be in the PATH, especially if its only used during build configuration (this is not made up, its a legitimate usecase). Can we not just do that, _and_ keep it in `tools.bzl`?

Szelethus (changes requested):
This can of worms has already consumed much more development time than it is worth. I had some success reproducing this or similar solutions on my environment, but I heard @nettle could not.

Lets abandon this patch but not the issue. I just wouldn't make this a priority considering its suprisingly low ROI.

PR #43, opened 2025-08-04
Add a single foss project to test rules on in ci

furtib:
Why: We want to test our rules on real-world examples
What: Added a single FOSS project to the CI pipeline

see issue: #20 

Note:
The goal is to build a robust test suite in small steps.

Depends on:
#62

PR #72, opened 2025-09-03
Create CI job for FOSS tests dynamically

furtib:
Why:
We want to test our rules on real-world examples.
We want this testing framework to be easily extendable.

What:
Added a CI job to collect projects and run tests on all of them.

Depends on:
#43

Addresses:
Fixes: #84

### DIFF
--- a/.github/workflows/foss.yaml
+++ b/.github/workflows/foss.yaml
@@ -74,7 +74,7 @@ jobs:
         
       - name: Run Per-File Bazel CodeChecker
         working-directory: ${{ matrix.project.folder }}/test-proj
-        run: bazel build :code_checker_test
+        run: bazel build :per_file_test
 
   foss_rhel_test:
     name: "Test rules on FOSS project: ${{ matrix.project.name }} (RHEL)"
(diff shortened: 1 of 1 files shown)


.github/foss_project_collector.sh
```diff
@@ -0,0 +1,39 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+if [ -z "$PATCH_DIR" ]; then
+    # PATCH_DIR not set
```
Szelethus, line 19: Might as well add an error message.

.github/foss_project_collector.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/env bash
```
Szelethus, line 1: Does this _need_ to be a shell script? I can't scratch the itch that this is much more complicated than it needs to be, considering that we have strict rules on how to add a new project to the pool. It might just be down to the fact that I find shell scripts just much harder to decypher against a python script (even if its a few lines lengthier).

Szelethus (commented):
Aha, so this is a best-practice solution detailed in https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations. I think we should highlight that in the comments somewhere.

Issue #20, opened 2025-07-24
Feature request: CI testing on opensource projects


furtib:
We should have a CI job test the code on smaller open-source projects, to test on real-world examples.

The idea as a template is viewable in: #16

# Scope:

1. Add CI job to run one, singular test
2. Expand the CI job to handle multiple tests
3. Add local running of tests
    - Add .gitignore for artifacts created
4. Document how to add tests to the test suite# DoD:

We have a working test set of at least 2 open source projects that can be run both locally and in github action

furtib:
We spoke about how the file structure is not good.

Currently, it is as follows:
- The main workflow (test.yml) is not an ideal name, but I have no idea what we could call it
- The install dependencies step is in actions/env_setup/action.yml. This was done because it is used in multiple tests. It is a composite action; the file name came from the GitHub guide I was following. I haven't paid much attention to it. Maybe we could move it to workflows/composit_actions/env_setup.yml
- The scripts that set up the projects to be tested currently reside in workflows/patches/ I thought I would call them patches because they change the BUILD and WORKSPACE files, but we could also call the folder init-scripts (?). Currently, the script file names follow this naming convention: patch-project_name.sh. This could be changed to just project_name.sh

PR #73, opened 2025-09-03
Foss local testing

furtib:
Why:
We want to test on open source projects. We want to be able to run the test locally.

What:
-  Added unittest to run all tests on all open-source projects.
- Updated gitignore to ignore tmp folders of FOSS projects.
- Blocked pytest from finding tests in the cloned projects.

Addresses:
Depends on:
#43

Addresses:
#20


test/foss/test_foss.py
```diff
@@ -0,0 +1,111 @@
...
+    return dirs
+
+
+PROJECT_DIRS = get_dynamic_test_dirs()
+
+
+# This will contain the generated tests.
+# I have not used the common lib from unit test, because it would
+# greatly increase the difficulty of the implementation, and this way the
+# two test aren't bound to each other
```
Szelethus, line 42: Don't use first person. Can't we just put the common library under `test` instead of `test/unit`? My understanding is that CodeChecker does that.

PR #76, opened 2025-09-08
Ci foss single threaded

furtib:
Why:
We want to test our rules on open-source projects in the CI. We want it fast and simple.

What:
Runs the local test runner in CI with pytest.

Addresses:
#20

Depends on: #73


furtib:
The only changed file is: foss.yaml

Szelethus (approved):
LGTM!

PR #78, opened 2025-09-08
Add test for compile flag filtering

furtib:
Why:
Add unit test for #19 

What:
Adds a unit test to check if `cxxopts` and `conlyopts` flags gets applied to their respective files

### DIFF
--- a/test/unit/compile_flags/test_compile_flags.py
+++ b/test/unit/compile_flags/test_compile_flags.py
@@ -54,30 +54,28 @@ class TestBasic(TestBase):
         with open(compile_commands) as f:
             json_content = json.load(f)
             for source in json_content:
-                # FIXME: Remove 'not' from condition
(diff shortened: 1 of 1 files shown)


test/unit/compile_flags/test_compile_flags.py
```diff
@@ -0,0 +1,101 @@
...
+            self.BAZEL_BIN_DIR,
+            "code_checker_filter",
+            "data",
+            "compile_commands.json",
+        )
+        self.assertTrue(os.path.exists(compile_commands))
+        with open(compile_commands) as f:
+            json_content = json.load(f)
+            for source in json_content:
+                # FIXME: Remove not from condition
```
Szelethus, line 86:
```suggestion
                # FIXME: Remove 'not' from condition
```

furtib:
I think I have found an even better one. Will merge once the test finishes.

PR #79, opened 2025-09-12
Generate metadata in distributed rule

### LINKED ISSUE #45: Generate metadata.json in file-by-file rule
The per-file CodeChecker Bazel rule doesn't generate metadata.json files. This is because every CodeChecker process generates its own metadata.json. [See here](https://github.com/furtib/codechecker_bazel/tree/metadata)

I'm not sure how to proceed with this.

We could, in theory, merge the contents of these metadata files into one.
(
This would entail:
- concat `"result_source_files"`
- For each analyzer: `"analyzer_statistics"`
    - add numbers in failed / successful
    - concat failed/successful_sources 
- add number in `"skipped"`
- update `"timestamps"` by setting begin to the lowest, end to the highest value
- For everything else, assume they are the same (All analysis should run with the same checkers enabled (?))

)

# Scope

-- Not sure

# DoD

Per-file Codechekcer Bazel rule generates a single file named `metadata.json`

furtib:
Why:
We currently do not produce a `metadata.json` with our distributed rule.

What:
Added metadata output for all steps.

Addresses:
#45

### DIFF
--- a/src/BUILD
+++ b/src/BUILD
@@ -5,6 +5,12 @@ py_binary(
     visibility = ["//visibility:public"],
 )
 
+py_binary(
+    name = "metadata_merge",
+    srcs = ["metadata_merge.py"],
+    visibility = ["//visibility:public"],
+)
+
 # Build & Test script template
 exports_files(
     ["codechecker_script.py"],
--- a/src/BUILD
+++ b/src/BUILD
@@ -5,15 +5,12 @@ py_binary(
     visibility = ["//visibility:public"],
 )
 
-py_binary(
-    name = "metadata_merge",
-    srcs = ["metadata_merge.py"],
-    visibility = ["//visibility:public"],
-)
-
 # Build & Test script template
 exports_files(
-    ["codechecker_script.py"],
+    [
+        "codechecker_script.py",
+        "metadata_merge.py",
+    ],
 )
 
 # The following are flags and default values for clang_tidy_aspect
--- a/src/metadata_merge.py
+++ b/src/metadata_merge.py
@@ -90,7 +90,7 @@ def save_json_file(data: Dict[str, Any], output_path: str) -> None:
         print(f"An error occurred while saving the file: {e}", file=sys.stderr)
 
 
-if __name__ == "__main__":
+def main():
     output_file = sys.argv[1]
     input_files = sys.argv[2:]
 
@@ -99,3 +99,6 @@ if __name__ == "__main__":
         save_json_file(merged_data, output_file)
     else:
         print("\nNo data was merged. Output file will not be created.")
+
+if __name__ == "__main__":
+    main()
--- a/src/metadata_merge.py
+++ b/src/metadata_merge.py
@@ -1,4 +1,17 @@
 #!{PythonPath}
+# Copyright 2023 Ericsson AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
 import json
 import os
--- a/src/metadata_merge.py
+++ b/src/metadata_merge.py
@@ -24,6 +24,8 @@ def merge_two_json(json1, json2):
         return json2
     if json2 == {}:
         return json1
+    # Fail if the plist file version is different
+    assert(json1["version"] == json2["version"])
     json1_root = json1["tools"][0]
     json2_root = json2["tools"][0]
     # We append info from json2 to json1 from here on out
--- a/src/metadata_merge.py
+++ b/src/metadata_merge.py
@@ -1,0 +1,86 @@
+import json
+import os
+import sys
+from typing import List, Dict, Any
+
+
+def merge_two_json(json1, json2):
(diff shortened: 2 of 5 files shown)


Szelethus (changes requested):

src/metadata_merge.py
```diff
@@ -0,0 +1,86 @@
...
+            json.dump(data, f, indent=4)
+        print(f"\nSuccessfully saved merged data to '{output_path}'")
+    except Exception as e:
+        print(f"An error occurred while saving the file: {e}", file=sys.stderr)
+
+
+if __name__ == "__main__":
+    output_file = sys.argv[1]
+    input_files = sys.argv[2:]
+
```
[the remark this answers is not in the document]
furtib: ~~This could be a good use of the common lib mentioned in #81 (fail, warn)~~
I added an assert for matching version numbers.

Szelethus (changes requested):
We need some tests to show the observable behaviour change this patch introduces, or patches this is a part of.

src/BUILD
```diff
@@ -23,6 +23,16 @@ py_binary(
     visibility = ["//visibility:public"],
 )
 
+# In bazel 6 we use our own python toolchain,
```
Szelethus: Which we no longer support ;)

furtib: How about this?
I reworded it to:
```python
# For targets using WORKSPACE files we use our own python toolchain,
# therefore we cannot load from rules_python.
```

src/metadata_merge.py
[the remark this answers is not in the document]
Szelethus: You are checking whether metadata files are mismatched, which is fine, but we should check whether the version is _exactly 2 or not_. If we see a new version, our merging algorithm likely needs fixing.

Szelethus (changes requested):
I have some thoughts. First, we add two new features:
 
* Enable metadata generation
* Add metadata merging

Can we somehow split this PR along that line? Also, with regards to testing, it would be great to have a test where we have two pre-existing metadata files which we merge using the new utility, and thats it.

src/per_file_script.py
```diff
@@ -157,7 +158,7 @@ def _display_error(ret_code: int) -> None:
 def _move_plist_files():
```
Szelethus: Don't we want to merge the new function with this one? They are essentially a copy-paste of one another. We could also hard assert whether metadata exists if the plist does.

Szelethus (changes requested):
Alright, looking good! Do you mean to land https://github.com/Ericsson/rules_codechecker/pull/254 before this? Can you rebase on it then?

src/per_file_script.py
furtib: Originally, I copied the metadata in that function, but changed it.
I don't remember why. I will take a look.

furtib: I merged the two.

Added an assert on the case when plist are generated, then metadata must be too.

src/per_file_script.py
```diff
@@ -181,22 +183,38 @@ def _move_plist_files():
                 shutil.move(
                     os.path.join(DATA_DIR, file_path), destination_plist_path
                 )
+                plist_exists = True
                 break
         else:
             with open(destination_plist_path, "w", encoding="utf-8") as file:
                 file.write(EMPTY_PLIST)
 
+    if os.path.isfile(os.path.join(DATA_DIR, "metadata.json")):
```
Szelethus, line 202: I find the docs here a little too loosy-goosy with regards to how much we're breaking the standard output of CodeChecker.

I'd expect something like this here, or somewhere in this function:

"A CodeChecker-compliant result directory for the *entire* analysis may have any number of plist files, but exactly one metadata.json file, as described in https://github.com/Ericsson/codechecker/blob/master/docs/report_directory.md. The problem is that in per-file mode, each translation unit is analyzed as a standalone analysis, each will have its own result directory and metadata.json file. To remain complaint, we will eventually merge all metadata files into a single one, but for now, we create a unique metadata file name before copying it over."

src/per_file_script.py
```diff
@@ -181,22 +183,38 @@ def _move_plist_files():
...
+        shutil.move(os.path.join(DATA_DIR, "metadata.json"), METADATA_FILE)
+    elif plist_exists:
+        # We do not create the "empty" metadata file, to make bazel fail
```
Szelethus: to make it fail? What does that mean?

Szelethus (changes requested):
The test looks great, thanks! Sorry for the comments in this round, could've made it earlier.

src/per_file_script.py
Szelethus: Also, if we are "making it fail", the warning below is actually an error.

src/per_file_script.py
furtib: done

src/per_file_script.py
furtib: I rephrased the comment

Szelethus (approved):
LGTM! Sorry for being so picky with this one, but I think the PR looks amazing now.

src/per_file_script.py
```diff
@@ -181,22 +183,49 @@ def _move_plist_files():
...
+
+    if os.path.isfile(os.path.join(DATA_DIR, "metadata.json")):
+        shutil.move(os.path.join(DATA_DIR, "metadata.json"), METADATA_FILE)
+    elif plist_exists:
+        # We do not create the "empty" metadata file
+        # this will cause bazel to crash due to a missing output file.
+        print(
+            "[WARNING] metadata.json doesn't exists "
+            "despite successful analysis..."
+        )
```
Szelethus, line 207: Sorry, final comment. Isn't this a should-never-happen case? If so, isn't this severe enough to warrant throwing an exception?

furtib: Done

Issue #31, opened 2025-07-29
Merge and/or rename code_checker_test


Szelethus:
In time, we should make sense of `code_checker_test`. We should likely rename it to something that alludes to the fact that it does per-file analysis as opposed to the "standard CodeChecker call" `codechecker_test` does. Over the long term, this approach should be the default, and it wouldn't make sense to keep them separate indefinitely.

[One proposed idea](https://github.com/Ericsson/codechecker_bazel/pull/13#discussion_r2164862933) by @nettle (keeping in mind that there already is industrial use of `codechecker_test` and we shouldn't mess with the API much) to add a `per_file` argument to `codechecker_test` which would actually trigger what `code_checker_test` does today as a standalon rule.

[Another suggestion](https://github.com/Ericsson/codechecker_bazel/pull/13#discussion_r2164866124), that even if we won't keep the separate public rule, we might keep the separate file. That should be named from `code_checker.bzl` to something like `per_file.bzl`.

PR #82, opened 2025-09-15
Rename code_checker_test to per_file, run it throught the codechecker_test macro

furtib:
Why:
We want a seamless migration from legacy users who used the monolithic solution.

What:
Extended the Macro invoking the codechecker_test rule with a new argument and redirected the rule to the distributed rule if that argument is set

Addresses:
#31

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,30 @@ codechecker_test(
 )
 ```
 
+#### Per-file CodeChecker analysis:
+> [!IMPORTANT]
+> The option is still in prototype status and is subject to changes or removal without notice. See [#31](https://github.com/Ericsson/codechecker_bazel/issues/31).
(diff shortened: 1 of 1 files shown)


Szelethus (commented):
I wonder if there is a way to deprecate a rule in bazel. Though I suppose we are not in a big rush to get rid of `code_checker_test`, we have plenty of internal-only rules.

Szelethus (commented):
Mostly LGTM, but I'd be happy to see a PR that showcases actually using the new parameter.

furtib:
See #92 .
It rewrites all tests to use the `codechecker_test` rule with the `per_file = True` argument.

Szelethus (approved):
Sick. Send it.

Issue #86, opened 2025-09-16
On the per_file rule default arguments and user given arguments are not merged correctly


furtib:
**Describe the bug**
On the `per_file` rule, default arguments and user-given arguments are not merged. They both get passed along to CodeChecker. This could be confusing users looking at the `codechecker.log`. Also, in a future patch, I want to depend on this argument list, and it's easier to filter them out here.

**Versions**
* `CodeChecker 6.26.0`
* `bazel 6.5.0`
* f4964166553f842b3aad688ee27468bd021b97e8

**To Reproduce**
Run in codechecker_bazel root:
(This will change the code_checker_pass rule!!!)
```bash
sed -i '262i\        "--analyzers clangsa clang-tidy",' test/unit/legacy/BUILD
bazel build //test/unit/legacy:code_checker_pass
cat bazel-bin/test/unit/legacy/code_checker_pass/data/test-unit-legacy-src-pass.cc_codechecker.log
```
See output:
```
CodeChecker command: CodeChecker analyze --analyzers clangsa clang-tidy --clean --ctu --analyzers clang-tidy clangsa cppcheck --output=code_checker_pass/data --file=*/test/unit/legacy/src/pass.cc bazel-out/k8-fastbuild/bin/test/unit/legacy/code_checker_pass/data/compile_commands.json.abs
...
```
**Expected behaviour**
I expected to see only one `--analyzers` in the CodeChecker command. There is 2.

PR #87, opened 2025-09-16
Add test for CodeChecker argument merging

furtib:
Why:
The default options and the user-defined ones don't get merged; they are just mashed together.

What:
Added a test that counts how many times --analyzers is defined

Addresses:
#86

### DIFF
--- a/test/unit/argument_merge/BUILD
+++ b/test/unit/argument_merge/BUILD
@@ -1,0 +1,40 @@
+# Copyright 2023 Ericsson AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


test/unit/argument_merge/BUILD
```diff
@@ -0,0 +1,40 @@
...
+)
+
+cc_library(
+    name = "main",
+    srcs = ["main.cc"],
+)
+
+# Test with overwriting analyzers argument
+code_checker_test(
+    name = "code_checker_argument_merge",
```
Szelethus, line 33: Keep in mind that we are just about to remove this.

Szelethus (approved):
LGTM with a nit.

Issue #65, opened 2025-08-26
Resolve all file paths to original in plist files


furtib:
### Describe the task

Currently, in multiple rules, the generated plist files point to the symlinks in the sandbox environment, not the original files. This could cause unexpected behaviours when parsing the output.

### Steps to reproduce

- Found in `test/unit/virtual_include/test_virtual_include.py`, look for these:

```lang=python
        # FIXME: In the postprocessed plists, all _virtual_include paths
        # should've been removed. Possible fix is in the github PR #14.
        self.assertNotEqual(
            self.contains_in_files(r"/_virtual_includes/", plist_files), []
        )
```

### Scope

- Fix file resolution implementation in the codechecker rule
- Implement file resolution in code_checker
- check the Clang static analyzer and clang tidy rules

### Definition of Done

- Have fixes for each
- Have a unittest for verifying paths being resolved for each rule

furtib:
## Solution Idea no. 4

We copy/link the files to the path where the list files would expect them to be.

As I understand, `CodeChecker store` uploads copies of the necessary files to the server, so it should not matter that the files we point to are at their original location or not.

### Cons:
- Maybe harder for the user to locate their files on the CodeChecker server (?)

PR #97, opened 2025-10-13
Add test for remote worker path resolution

furtib:
Why:
We want to test our solutions for making remote workers' absolute paths into locally usable relative paths. We should add a test for it.

What:
Adds a new test to run on pre-created remote worker paths.

Addresses:
#65

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,6 +1,6 @@
 [MASTER]
 # To find common lib for tests
-init-hook='import sys; sys.path.append("test")'
+init-hook='import sys; import os; sys.path.append("test"); sys.path.append(os.getcwd())'
 ignore = test_template.py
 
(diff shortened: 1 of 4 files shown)


Szelethus (changes requested):
I'm fine with creating/removing the tmp directory with a class level setup/teardown, but by having a global `PATH_RESOLUTION`, you practically do 90% of the test in those methods. In fact, I'm not even sure the way you are testing there is a need for creating any files. I recommend that you simply test the string replace, which is all that really matters here.

Szelethus (commented):
Awesome, this is what I wanted to see! I have some nits, but otherwise this is a great PR.

Szelethus (approved):
LGTM, though I believe pylint will catch this. Please rebase.

PR #98, opened 2025-10-21
Use the compile_commands implementation from the monolithic rule in the per_file rule

furtib:
Why:
Currently, we use a different, inferior implementation for generating compile_commands.json in the `pre_file` rule.
We should use the same as in the monolithic rule. This would prevent us from having to maintain both rules simultaneously. 

What:
Replaces the `compile_commands.json` generating implementation in the `pre_file` rule.
Fixes the unittest for compile_commands, the per_file `compile_commands.json` location changed

Addresses:
Fixes: #120 
Fixes: #128

### DIFF
--- a/src/per_file.bzl
+++ b/src/per_file.bzl
@@ -57,9 +57,7 @@ def _run_code_checker(
         inputs = [compile_commands_json, config_file] + sources_and_headers
     else:
         # NOTE: we collect only headers, so CTU may not work!
-        headers = depset(direct = [],
-                         transitive = target[SourceFilesInfo].headers.to_list()
-                        )
+        headers = depset(transitive = target[SourceFilesInfo].headers.to_list())
         inputs = depset([compile_commands_json, config_file, src], transitive = [headers])
 
     outputs = [clang_tidy_plist, clangsa_plist, codechecker_log]


furtib:
Fails because #99 does not affect the "monolithic" compile command generator implementation. should rebase after merging #100

Szelethus (changes requested):
Just like that, and everything runs? Damn.

src/per_file.bzl
```diff
@@ -52,7 +57,11 @@ def _run_code_checker(
         inputs = [compile_commands_json, config_file] + sources_and_headers
     else:
         # NOTE: we collect only headers, so CTU may not work!
-        headers = depset([src], transitive = [compilation_context.headers])
+        trans_depsets = [compilation_context.headers]
+        for dset in target[SourceFilesInfo].headers.to_list():
+            trans_depsets.append(dset)
+        headers = depset(direct = [src],
+            transitive = trans_depsets)
```
Szelethus: Lets turn this into a function.

src/per_file.bzl
```diff
@@ -52,7 +57,11 @@ def _run_code_checker(
         inputs = [compile_commands_json, config_file] + sources_and_headers
     else:
         # NOTE: we collect only headers, so CTU may not work!
-        headers = depset([src], transitive = [compilation_context.headers])
+        trans_depsets = [compilation_context.headers]
+        for dset in target[SourceFilesInfo].headers.to_list():
+            trans_depsets.append(dset)
+        headers = depset(direct = [src],
```
nettle: Not sure I understand this line...
So, this is not just headers but src as well?

src/per_file.bzl
```diff
@@ -52,7 +57,11 @@ def _run_code_checker(
         inputs = [compile_commands_json, config_file] + sources_and_headers
     else:
         # NOTE: we collect only headers, so CTU may not work!
-        headers = depset([src], transitive = [compilation_context.headers])
+        trans_depsets = [compilation_context.headers]
+        for dset in target[SourceFilesInfo].headers.to_list():
```
nettle: I guess this is a header :)

src/compile_commands.bzl
```diff
@@ -418,6 +418,29 @@ def compile_commands_impl(ctx):
         ),
     ]
 
+def get_compile_commands_json_and_srcs(ctx):
```
nettle: Maybe I'm mistaken, at least I do not see that yet, but I guess adding this function does not give us much benefits:
As I mentioned I would prefer original implementation in src/codechecker.bzl.
And in src/per_file.bzl we even do not need this function.

nettle (changes requested):
May I suggest changing only src/per_file.bzl implementation first, which seems reasonable to me and not touching src/codechecker.bzl and src/compile_commands.bzl at least yet.

Szelethus:
> May I suggest changing only src/per_file.bzl implementation first, which seems reasonable to me and not touching src/codechecker.bzl and src/compile_commands.bzl at least yet.

It is a more than reasonable ask to land parts of this patch independently, but I'm strongly against not touching those two files to keep them stable, especially with NFC patches.

src/per_file.bzl
```diff
@@ -300,7 +142,9 @@ def _create_wrapper_script(ctx, options, compile_commands_json, config_file):
     )
 
 def _per_file_impl(ctx):
-    compile_commands_json = _compile_commands_impl(ctx)
+    compile_commands_json, source_files = \
+        get_compile_commands_json_and_srcs(ctx)
```
nettle: The following can be used to get compile_commands_json:
```
for output in compile_commands_impl(ctx):
    if type(output) == "DefaultInfo":
        compile_commands_json = output.files.to_list()[0]
```

nettle (changes requested):

src/per_file.bzl
furtib: Done, see: `_flatten_depset`

src/per_file.bzl
```diff
@@ -92,190 +111,22 @@ def check_valid_file_type(src):
...
-        if CompileInfo in target:
-            if hasattr(target[CompileInfo], "arguments"):
-                srcs = target[CompileInfo].arguments.keys()
+        if SourceFilesInfo in target:
+            if (hasattr(target[SourceFilesInfo], "transitive_source_files")
+            and hasattr(target[SourceFilesInfo], "headers")):
+                srcs = target[SourceFilesInfo].transitive_source_files.to_list()
+                headers = _flatten_depset(
+                    target[SourceFilesInfo].headers
+                    ).to_list()
```
nettle: And we kinda doing this again :)

src/per_file.bzl
```diff
@@ -17,15 +17,29 @@ Rulesets for running codechecker in a different Bazel action
...
-load("common.bzl", "SOURCE_ATTR")
+load(
+    "compile_commands.bzl",
+    "compile_commands_aspect",
+    "compile_commands_impl",
+    "platforms_transition",
+    "SourceFilesInfo",
+)
+load(
+    "@codechecker_bazel//src:codechecker_config.bzl",
```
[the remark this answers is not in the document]
furtib: This line was untouched in this PR; editing it is not in the scope of this patch.

I always wondered why it is that sometimes we use the workspace prefix, sometimes we don't.
Does this have any significance, or is it just purely aesthetic?

src/per_file.bzl
```diff
@@ -49,7 +70,17 @@ def _run_code_checker(
...
-        headers = depset([src], transitive = [compilation_context.headers])
+        headers = depset(direct = [],
+                         transitive = [
+                             compilation_context.headers,
+                             # This is necessary because
+                             # target[SourceFilesInfo].headers contains
+                             # depsets of header files, and we can only
+                             # append depsets of files not depset of depsets
+                             # to the headers depset.
+                             _flatten_depset(target[SourceFilesInfo].headers)
```
nettle: I just wonder why it works in case of compile_commands_impl?

src/per_file.bzl
nettle: I mean we collect all headers here again (extracting them from probably buggy structure). I think we do not need _flatten_depset() function, but _get_all_headers() function?

src/per_file.bzl
```diff
@@ -294,7 +145,14 @@ def _create_wrapper_script(ctx, options, compile_commands_json, config_file):
     )
 
 def _per_file_impl(ctx):
-    compile_commands_json = _compile_commands_impl(ctx)
+    compile_commands = None
+    for output in compile_commands_impl(ctx):
+        if type(output) == "DefaultInfo":
+            compile_commands = output.files.to_list()[0]
```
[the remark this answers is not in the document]
nettle: Never mind :)

Issue #99, opened 2025-11-12
Generated source files aren't getting analysed


furtib:
**Describe the bug**
In case of generated files (tested with genrule(), but should apply to every generation method)

**Versions**
* `CodeChecker 6.26.0`
* `bazel 6.5.0`
* ff788bfc7f1bd6f6a9b79cd4a72ffedc2612e284

**To Reproduce**
Run this script and run `bazel build //app:per_file_rule`, observe that the `foo.cpp` doesn't get analyzed!
```bash
mkdir present-bug
cd present-bug
cat > WORKSPACE <<'endtag'
#----------------------------------------------------

load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")

git_repository(
    name = "bazel_codechecker",
    remote = "https://github.com/Ericsson/codechecker_bazel.git",
    branch = "main",
)

load(
    "@bazel_codechecker//src:tools.bzl",
    "register_default_codechecker",
    "register_default_python_toolchain",
)

register_default_python_toolchain()

register_default_codechecker()

#----------------------------------------------------
endtag

mkdir app

cat > app/BUILD <<'endtag'
#-------------------------------------------------------

# codechecker rules
load(
    "@bazel_codechecker//src:codechecker.bzl",
    "codechecker_test",
)
load(
    "@bazel_codechecker//src:per_file.bzl",
    "per_file_test",
)


codechecker_test(
    name = "codechecker_test",
    targets = [
        ":hello_app",
    ],
)

per_file_test(
    name = "per_file_test",
    targets = [
        ":hello_app",
    ],
    options = ["--ctu"]
)

#-------------------------------------------------------

genrule(
    name = "create_header",
    # This is the file we will generate.
    outs = ["generated_message.h"],
    cmd = "echo 'static const char* MESSAGE = \"Hello from a generated file!\";\nint asd = 0;' > $(location generated_message.h)",
)

genrule(
    name = "create_foo",
    outs = [
        "gen_foo.h",
        "gen_foo.cpp",
    ],
    cmd = "echo 'int foo();' > $(location gen_foo.h) && echo 'int foo() { return 0; }' > $(location gen_foo.cpp)",
)
cc_binary(
    name = "hello_app",
    srcs = [
        "main.cpp",
        ":generated_message.h",
        ":gen_foo.h",
        ":gen_foo.cpp",
    ],
)
endtag


cat > app/main.cpp <<'endtag'
#include <iostream>

[...]

PR #100, opened 2025-11-12
Support genfile rules

furtib:
Added generated files to the CompileInfo aspect.

Why:
We want to support analysis on generated files, so that they can be used during CTU analysis

What:
Removed is_source validation for source files.
Generated files are regarded as output, not input, and thus they aren't regarded as source files and were filtered from the list of source files.

Addresses:
#99 
Depends on: #101


nettle:
https://bazel.build/rules/lib/builtins/File#is_source

PR #102, opened 2025-11-24
Extend config handling to yaml files

### LINKED ISSUE #30: Add support for YAML CodeChecker config files
As discussed in the documentation patch #13, we [likely also support](https://github.com/Ericsson/codechecker_bazel/pull/13#discussion_r2164858937) YAML config files as well. Here is the corresponding CodeChecker documentation: https://github.com/Ericsson/codechecker/blob/master/docs/config_file.md#yaml.

My initial impression is to the contrary. I created config.yaml to the image of config.json already in the repo:

**config.json**:
```json
{
    "analyze": [
        "--enable=core.DivideZero",
        "--enable=core.CallAndMessage",
        "--analyzer-config"
    ],
    "parse": [
        "--print-steps"
    ],
    "server": [
        "--port=9090"
    ],
    "store": [
        "--name=run_name",
        "--tag=tag_name"
    ]
}
```
**config.yaml**:
```yaml
---
analyze:
- "--enable=core.DivideZero"
- "--enable=core.CallAndMessage"
- "--analyzer-config"
parse:
- "--print-steps"
server:
- "--port=9090"
store:
- "--name=run_name"
- "--tag=tag_name"
```

In order to test whether this could be a drop-in replacement, I changed one of the checker names to something nonexistent, which should trigger a CodeChecker error. In the json version, it correctly did after invoking `bazel test codechecker_pass`. However, when I changed the filegroup to take the yaml file instead of the json one:

```starlark
filegroup(
    name = "codechecker_config_file",
    srcs = [":config.yaml"],
)
```

calling `bazel test codechecker_pass` didn't give me an error, even though the config gile had an erronous checker name.

This should be investigated further, it shouldn't be too tricky to also support YAML config files.

furtib:
Why:
We want to support not just JSON, but YAML config files too.

What:
- Added support for YAML config files in the monolishic codechecker rule.
- Added support for any config file in the per_file rule

Addresses:
Fixes #30
Depends on: #103

### DIFF
--- a/src/per_file_script.py
+++ b/src/per_file_script.py
@@ -91,7 +91,7 @@ def _run_codechecker() -> None:
         + ["--config=" + CONFIG_FILE]
         + [COMPILE_COMMANDS_ABSOLUTE]
     )
-    log(f"CodeChecker command: {codechecker_cmd}\n")
+    log(f"CodeChecker command: {' '.join(codechecker_cmd)}\n")
     log("===-----------------------------------------------------===\n")
     log("                   CodeChecker error log                   \n")
     log("===-----------------------------------------------------===\n")


furtib:
Superseeded by #113, #112 and #104

PR #103, opened 2025-11-24
Create Test case for config files

furtib:
Why:
We want to support config files; we should test for them.
What:
- Test whether every rule that takes a config file as an argument acts on the content of said config file 

Addresses:
#30
Test for #102

### DIFF
--- a/test/unit/config/test_config.py
+++ b/test/unit/config/test_config.py
@@ -72,7 +72,7 @@ class TestConfig(TestBase):
             "codechecker_test_json",
             "config.json"
         )
-        # After the path the file name will change
+        # After the fix the file name will change
         # from codechecker_config.json to config.json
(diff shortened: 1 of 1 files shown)


test/unit/config/test_config.py
```diff
@@ -0,0 +1,129 @@
...
+            "bazel test //test/unit/config:codechecker_test_json"
+        )
+        # Should not find the division by zero bug
+        self.assertEqual(ret, 0)
+        copied_config = os.path.join(
+            self.BAZEL_BIN_DIR, # type: ignore
+            "codechecker_test_json",
+            "config.json"
+        )
+        # After the path the file name will change
```
Szelethus: I suppose you mean patch, not path, and I think we should phrase it as "Ideally after a fix, the file name ..."

PR #104, opened 2025-11-25
Refactor config rule

furtib:
Why:
We want to handle configurations for both the monolithic and `per_file` rule.
I want to use the same implementation for both, and don't want to either repeat code or have cyclic includes.

What:
I have created a file `codechecker_config.bzl` in which everything related to the rule codechecker_config lives.
I have included this rule in `codechecker.bzl` and used it.

Addresses:
-
(In the long run: #30 )

### DIFF
--- a/src/codechecker.bzl
+++ b/src/codechecker.bzl
@@ -371,6 +371,11 @@ def codechecker_suite(
         **kwargs
     )
 
+# This rule definition is here for compatibility reasons
+# in the earliest versions, the entire codechecker_config definition was here,
+# but was later moved to its own .bzl file.
+# This macro is left here so that early adopters
+# don't need to change where the rule is loaded from.
 def codechecker_config(
         name,
         analyze = [],


src/codechecker_config.bzl
```diff
@@ -0,0 +1,124 @@
...
+            config_info = ctx.attr.config[0][CodeCheckerConfigInfo]
+        else:
+            config_info = ctx.attr.config[CodeCheckerConfigInfo]
+        if config_info.config_file:
+            # Create a copy of CodeChecker configuration file
+            # provided via codechecker_config(config_file)
+            config_file = config_info.config_file.files.to_list()[0]
+            config_file_name = ctx.attr.name + \
+                                "/config." + config_file.extension
+    ctx_config_file = ctx.actions.declare_file(config_file_name)
```
furtib: Note: This part is not the same as it was in `codechecker.bzl`

Issue #108, opened 2025-11-27
Support newer Bazel versions


furtib:
### Describe the task

All rules should work flawlessly under newer Bazel versions (Especially 8.X.X).

### Scope

- Support Bazel 7
- Support Bazel 8

### Definition of Done

We support at least Bazel 8, and if possible, also Bazel 7 and 6.

PR #111, opened 2025-12-01
Support bazel 7

furtib:
Why:
We want to extend our support for newer Bazel versions

What:
Added a variable `BAZEL_VERSION` to the generated `defs.bzl`
Made the attribute `_whitelist_function_transition` conditional. It is only added to the attributes list if the user uses a Bazel version older than 6.

Addresses:
#108

Note:
The dictionary concatenation used may have been introduced in Bazel 6

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Prerequisites
 We need the following tools:
 
 - Git 2 or newer (we use 2.36)
-- Bazel 6, not yet bazel 7 (we recommend version 6.5.0)
+- Bazel 6 or 7, not yet bazel 8 (we recommend version 7.7.0)
 - Clang 16 or newer (we use 16), we use clang-tidy
 - Python 3.8 or newer (we use 3.11)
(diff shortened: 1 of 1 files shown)


furtib:
Sorry, that comment was a bit outdated.
I have already started working on a more robust GitHub action suite so that it can be easily extended to multiple Bazel versions.
see #115

src/common.bzl
```diff
@@ -0,0 +1,12 @@
+load(
+    "@default_codechecker_tools//:defs.bzl",
+    "BAZEL_VERSION"
+)
+
+def old_bazel_attributes():
```
Szelethus: Please add some comments describing what this is and why we need it.

Szelethus (approved):
Alright, lets do it. Since the github CI changes land here at the same time, we can't really move the test changes to a separate PR, so this is as lean as this patch can get.

PR #115, opened 2025-12-02
Prepare unittests to be extended with different bazel versions

furtib:
Why:
We want to support multiple Bazel versions simultaneously.
We should run our test suite on all supported Bazel versions.

What:
Added a matrix strategy to both of the unittest jobs. `.github/bazel_version.json` contains a list of bazel versions to be used during testing.

Addresses:
#108


furtib:
I have gone ahead with option A

Szelethus (approved):
This looks like a very elegant solution! LGTM!

PR #116, opened 2025-12-02
Showcase Bazel 8 support

furtib:
Why:
We want to support the latest bazel version

What:
Waaaay too much.
Only the codechecker_test rule have been updated to support bazel 8
Note: the MODULE.bazel is based on an older implementation of #110, this is why files like: extensions.bzl exists

Addresses:
#108

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,0 +1,20 @@
+"""
+Module docs string
+"""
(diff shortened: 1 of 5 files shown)


furtib:
Outdated, see #140 instead

PR #119, opened 2025-12-16
Add license to tests that are missing it

furtib:
Why:
All files should start with a license comment.

What:
Added the license to all test files missing it

Addresses:
none


Szelethus (approved):
LGTM

Issue #120, opened 2026-01-05
Missing header files during analysis


furtib:
**Describe the bug**
On targets with dependencies using strip_include_prefix, header files are not found correctly.

**Versions**
* `CodeChecker 6.26.1`
* `bazel 6.5.0`
* 2f291d3da07987d41e28e2447c9c7b0a0ccc03dd

**To Reproduce**
The following script creates a project that compiles, but cannot be analyzed.
```bash
mkdir present-bug
cd present-bug
cat > WORKSPACE <<'endtag'
#----------------------------------------------------

load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")

git_repository(
    name = "bazel_codechecker",
    remote = "https://github.com/Ericsson/codechecker_bazel.git",
    branch = "main",
)

load(
    "@bazel_codechecker//src:tools.bzl",
    "register_default_codechecker",
    "register_default_python_toolchain",
)

register_default_python_toolchain()

register_default_codechecker()

#----------------------------------------------------
endtag

cat > BUILD <<'endtag'
#-------------------------------------------------------

# codechecker rules
load(
    "@bazel_codechecker//src:codechecker.bzl",
    "codechecker_test",
)
load(
    "@bazel_codechecker//src:per_file.bzl",
    "per_file_test",
)


codechecker_test(
    name = "codechecker_test",
    targets = [
        ":hello_app",
    ],
)

per_file_test(
    name = "per_file_test",
    targets = [
        ":hello_app",
    ],
    options = ["--ctu"]
)

#-------------------------------------------------------

cc_library(
    name = "hello_app",
    srcs = [
        "main.cpp",
    ],
    implementation_deps = ["//sub:dep_lib"],
)

endtag


cat > main.cpp <<'endtag'
#include <iostream>

#include "sub.h"

int main() {
    int y = 1/foo();
    int x = 1/bar();
    return 0;
}
endtag

mkdir sub

cat > sub/BUILD << 'endtag'
cc_library(
    name = "dep_lib",
    hdrs = ["sub.h"],
    srcs = ["sub.cpp"],
    strip_include_prefix = ".",
    visibility = ["//visibility:public"],
)
endtag

cat > sub/sub.h <<'endtag'
#pragma once
int foo(){
    return 0;
}
int bar();
endtag

cat > sub/sub.cpp <<'endtag'
#include <iostream>

#include "sub.h"

int bar() {
    int c = 0;
    int f = 0/c;
    return 0;
[...]

PR #121, opened 2026-01-05
Add test for missing header bug, under implementation_deps targets

furtib:
Why:
Header files under `implementation_deps` dependencies are hidden if the dependency library uses `strip_include_prefix`.

What:
Added tests to validate whether the bug has been fixed. 

Addresses:
#120


test/unit/virtual_include/BUILD
furtib: This is where strip_include_prefix is used.

PR #125, opened 2026-01-06
Update rules for clang-tidy and clang --analyze

nettle:
Why:
Bazel rules for clang-tidy and clang --analyze
were never updated and never tested on real projects.

What:
- Collect sources and headers in transitive dependencies
- Support platform transitions (for multi platform projects)
- Simplify launcher scripts, bazelify parameters
- Fix plist generation for clang --analyze
- Support config file to accept input from a file_group
- Remove clang-tidy aspect since it is not actually used
- Minor cosmetic changes
- Test on some of real Ericsson projects
- Updated README

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,8 @@ The main Bazel rule for CodeChecker is `codechecker_test()`.
 
 ### Clang-tidy
 
-Clang-tidy is a fast static analyzer/linter for the C family of languages. This
-repository provides Bazel aspect `clang_tidy_aspect()` and rule `clang_tidy_test()`
+Clang-tidy is a fast static analyzer/linter for the C family of languages.
+This repository provides Bazel rule `clang_tidy_test()`
 to run clang-tidy natively (without CodeChecker).
 
 Find more information about LLVM clang-tidy:
@@ -313,9 +313,10 @@ codechecker_test(
 
 The following rules are _not_ using CodeChecker.
 
-### Clang-tidy: `clang_tidy_aspect()` and `clang_tidy_test()`
+### Clang-tidy: `clang_tidy_test()`
 
-The Bazel rule `clang_tidy_test()` runs clang-tidy natively without CodeChecker. To use it, add the following to your BUILD file:
+The Bazel rule `clang_tidy_test()` runs clang-tidy natively without CodeChecker.
+To use it, add the following to your BUILD file:
 
 ```python
 load(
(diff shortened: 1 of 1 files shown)


src/clang.bzl
```diff
@@ -448,16 +413,31 @@ def _clang_test(ctx, tool):
...
+def _extra_attributes():
+    if BAZEL_VERSION.startswith("6."):
+        return {
+            "_whitelist_function_transition": attr.label(
+                default = "@bazel_tools//tools/whitelists/function_transition_whitelist",
+                doc = "needed for transitions",
+            ),
+        }
+    return {}
+
```
furtib: I think this should wait on #111.

src/clang.bzl
```diff
@@ -288,147 +299,101 @@ def _valid_for_clang_tidy(target, ctx):
...
-                if hasattr(dep[CompileInfo], "arguments"):
-                    srcs = dep[CompileInfo].arguments.keys()
-                    sources += srcs
-    return sources
-
-def _collect_all_sources(ctx):
-    sources = _rule_sources(ctx)
-    for attr in source_attr:
+def _process_all_deps(ctx, arguments, headers):
+    for attr in ["srcs", "deps", "data", "exports"]:
```
furtib: Lets use the global variable for these attributes,
For this load:
```
load(
    "@bazel_codechecker//src:tools.bzl",
    "source_attr"
)
```
```suggestion
    for attr in source_attr::
```

src/clang.bzl
```diff
@@ -485,7 +466,7 @@ clang_tidy_test = rule(
             cfg = "exec",
             doc = "Clang-tidy executable",
         ),
-    },
+    } | _extra_attributes(),
```
furtib: Use the solution from #111

src/clang.bzl
```diff
@@ -527,7 +521,7 @@ clang_analyze_test = rule(
             cfg = "exec",
             doc = "Clang executable",
         ),
-    },
+    } | _extra_attributes(),
```
furtib: Use the solution from #111

test/unit/legacy/test_legacy.py
```diff
@@ -130,22 +130,6 @@ def test_bazel_compile_commands(self):
         self.grep_file(compile_commands, r"pass\.cc")
         self.grep_file(compile_commands, r"\/gcc")
 
-    def test_bazel_aspect_clang_tidy_pass(self):
```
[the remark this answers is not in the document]
nettle: Separate implementation of aspect for clang-tidy is excessive, and we basically do not use it - it is confusing, a bit difficult to use (complex command line) and most importantly it does not support dependencies which makes it useless in some projects.

src/clang.bzl
nettle: OK, agree. The only thing I suggest is that global constants should be CAPITALIZED I guess

src/clang.bzl
```diff
@@ -288,147 +299,101 @@ def _valid_for_clang_tidy(target, ctx):
...
-    # headers = depset()
+    all_headers = depset()
     for target in ctx.attr.targets:
-        if not CcInfo in target:
-            continue
         if CompileInfo in target:
             if hasattr(target[CompileInfo], "arguments"):
                 srcs = target[CompileInfo].arguments.keys()
+                headers = target[CompileInfo].headers
+                # print(":::", target.label, ":", len(srcs), len(headers.to_list()))
```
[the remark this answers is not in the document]
nettle: Good suggestion!
Or... I'll just remove this, I just missed this print

src/clang.bzl
nettle: Done

Issue #128, opened 2026-01-09
Missing header files from external repositories during analysis


furtib:
**Describe the bug**
On targets with implementation_deps **external** dependencies using strip_include_prefix, header files are not found correctly.

**Versions**
Versions of `CodeChecker`, `bazel` and this repo. You can get version info via:
* `CodeChecker 6.26.1`
* `bazel 6.5.0`
* 2f291d3da07987d41e28e2447c9c7b0a0ccc03dd

**To Reproduce**
Checkout this branch: https://github.com/Szelethus/codechecker_bazel/tree/test_system_headers

**Expected behaviour**
External headers to be found.

**Additional context**
This is an edge case of #120

PR #129, opened 2026-01-10
Add a test for external dependencies that are also implementation_deps

Szelethus:
Why:
There is a funny corner case with implementation_deps, when the dependency is an external dependency as well. In that case, bazel constructs the compilation commands by including from said dependency with `-isystem` instead of `-I`, which we don't handle.

What:
The patch reproduces this issue with a test. Instead of having an _external_ external library, I created a local repository which achieves the same thing. Since I needed to create local repositories, I also needed to adjust the .gitignore file.

Addresses:
Related to #128

### DIFF
--- a/test/unit/external_repository/.bazelrc
+++ b/test/unit/external_repository/.bazelrc
@@ -1,1 +1,15 @@
+# Copyright 2023 Ericsson AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


furtib (approved):
LGTM,
With a nit, these licenses may be a bit too much. I don't have a strong opinion on having them (or not) in .bazelrc files.

Issue #124, opened 2026-01-06
Name inconsistency codechecker_bazel vs bazel_codechecker


nettle:
### Describe the task

There is a name inconsistency between repo name **codechecker_bazel** and workspace/module name **bazel_codechecker**.
I think we should use one, and I guess it should be **codechecker_bazel**.

### Scope

1. Scan the source, tests, documentation etc to reveal workspace/module and repo names
2. Estimate changes
3. Replace
4. Make sure all tests pass
5. Make sure documentation is consistent

### Definition of Done

Workspace/module name is the same as repo name: **codechecker_bazel**

PR #135, opened 2026-01-12
Rename bazel_codechecker to codechecker_bazel

furtib:
Why:
We want the rule name to be consistent with the repository name

What:
Replaced most occurrences of `bazel_codechecker` with `codechecker_bazel` (notable exceptions are test names)

Addresses:
Fixes: #124


Szelethus (approved):
Great idea to fix this up before the release. I expect this will break pretty much every independent CI tho.

PR #138, opened 2026-01-12
Reactivate passing of config parameter to per_file rule

furtib:
Why:
We forgot to activate this when merging Bazel 7 support.

What:
Uncomments the passing of the config parameter to the per file rule, through the codechecker_test macro. 

Addresses:
none


Szelethus (approved):
LGTM. Good thing that we merged the bazel 7 support _after_ the release.

PR #140, opened 2026-01-15
Replace python toolchain with @rules_python

furtib:
Why:
We should use an external Python toolchain instead of maintaining our own.
This is necessary mainly for Bazel 8 support.

What:
- Replaced the Python toolchain

Addresses:
#108

### DIFF
--- a/src/codechecker.bzl
+++ b/src/codechecker.bzl
@@ -233,7 +233,7 @@ def _codechecker_test_impl(ctx):
         output = ctx.outputs.codechecker_test_script,
         is_executable = True,
         substitutions = {
-            "#{Python_path}": py_toolchain.stub_shebang,
+            #"#{Python_path}": py_toolchain.stub_shebang,
             "{Mode}": "Test",
             "{Verbosity}": "INFO",
             "{codechecker_bin}": CODECHECKER_BIN_PATH,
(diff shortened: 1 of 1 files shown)


.github/platform_environment_setup/rhel9/action.yaml
```diff
@@ -41,6 +41,11 @@ runs:
           VERSION=1.26.0; \
           wget "https://github.com/bazelbuild/bazelisk/releases/download/v$VERSION/bazelisk-linux-amd64" && \
           chmod +x bazelisk-linux-amd64 && \
-          mv bazelisk-linux-amd64 /usr/local/bin/bazel && \
-          USE_BAZEL_VERSION=6.5.0 bazel version
+          mv bazelisk-linux-amd64 /usr/local/bin/bazel
+        shell: bash
+      
+      - name: Change to a non-root user
```
Szelethus: This is a very non-obvious thing to do. This needs some documentation.

src/codechecker.bzl
```diff
@@ -130,8 +132,8 @@ def _codechecker_impl(ctx):
             codechecker_files,
             ctx.outputs.codechecker_log,
         ],
-        executable = ctx.outputs.codechecker_script,
-        arguments = [],
+        executable = py_toolchain.interpreter,
+        arguments = [ctx.outputs.codechecker_script.path],
```
nettle: Maybe, maybe...

src/codechecker.bzl
```diff
@@ -196,10 +198,8 @@ codechecker = rule(
             default = ":codechecker_script.py",
             allow_single_file = True,
         ),
-        "_python_runtime": attr.label(
-            default = "@default_python_tools//:py3_runtime",
-        ),
     },
+    toolchains = ["@rules_python//python:toolchain_type"],
```
nettle: I was also thinking about toolchain, but for some reasons used runtime.
It is definitely good idea to try toolchain again!

src/codechecker.bzl
```diff
@@ -229,28 +226,36 @@ def _codechecker_test_impl(ctx):
...
         fail("Execution results required for codechecker test are not available")
 
+    py_toolchain = ctx.toolchains["@rules_python//python:toolchain_type"].py3_runtime
     # Create test script from template
     ctx.actions.expand_template(
         template = ctx.file._codechecker_script_template,
         output = ctx.outputs.codechecker_test_script,
         is_executable = True,
         substitutions = {
+            #"#{Python_path}": py_toolchain.stub_shebang,
```
nettle: This might be interesting!
I did not know about stub_shebang!

PR #141, opened 2026-01-15
Create a convenience function for interacting with compile_commands_impl

Szelethus:
Why:
In anticipation of #98, where we remove the compilation database generation from per_file analysis and reuse compile_commands.bzl, we identified that the best point of entry might not be `compile_commands_impl()`, because we still need to check and parse the return values. Since we need the exact same ~20 lines of code that codechecker.bzl already uses, lets just turn that into a convenience function.

What:
Turn the handling/parsing of compile_commands_impl into a function.

Addresses:
Accessory to #98.

### DIFF
--- a/src/codechecker.bzl
+++ b/src/codechecker.bzl
@@ -19,7 +19,7 @@ Rulesets for running codechecker in a single Bazel job.
 load(
     "compile_commands.bzl",
     "compile_commands_aspect",
-    "compile_commands_impl",
+    "get_compile_commands_json_and_srcs",
     "platforms_transition",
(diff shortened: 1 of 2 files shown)


furtib (approved):
LGTM

nettle (changes requested):
To be honest I'm kind of against this change.
May we discuss this before submitting?

nettle:
Hi @Szelethus,
What I meant is #145 - very simple function which can be used in #98 without any changes in src/codechecker.bzl
This is still very optional

PR #145, opened 2026-01-16
Convenience function for interacting with compile_commands_impl

nettle:
Why:
Simplify integration src/compile_commands.bzl into src/per_file.bzl

What:
- Introduce get_compile_commands() function which returns only compile_commands.json File object and performs optional checks

Addresses:
Optional accessory to #98

### DIFF
--- a/src/compile_commands.bzl
+++ b/src/compile_commands.bzl
@@ -418,6 +418,27 @@ def compile_commands_impl(ctx):
         ),
     ]
 
+def get_compile_commands(ctx, check=True):
+    """ Returns the compilation database file
+
+    Returns:
+      compile_commands.json File object
+    """
+    compile_commands = None
+    source_files = None
+    for output in compile_commands_impl(ctx):
+        if type(output) == "DefaultInfo":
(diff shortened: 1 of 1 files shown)


Szelethus (commented):
Besides the check parameter, this was copy-pasted from codechecker.bzl. Please help me understand what specific concerns you have about not reusing this code?

Again, this is not a hill I intend to die on, and I'm fine to support a solution I understand, even if I disagree with it.

nettle:
> Besides the check parameter, this was copy-pasted from codechecker.bzl. Please help me understand what specific concerns you have about not reusing this code?
> 
> Again, this is not a hill I intend to die on, and I'm fine to support a solution I understand, even if I disagree with it.

I actually do not see much use of this #145 as well as #141 :)
I implemented this convenience function in a form that might be used in per_file.bzl
But, as to me, it is not needed

Szelethus:
This does something similar to https://github.com/Ericsson/rules_codechecker/pull/141, and even that is in backlog. I think its fine to close of the two.

PR #146, opened 2026-01-16
Use python toolchain instead of python runtime

nettle:
Why:
Need to make python runtime more portable
and be ready for rules_python support.

What:
- Add toolchain for python instead of _python_runtime
- Add portable functions for python
- Fix register_toolchains

Addresses:
First step for rules_python support (Bazel 8) #108


furtib (changes requested):
This patch looks great!
The testing breaks only because the WORKSPACE changed!
This change should also be represented in `README.md`.

Issue #148, opened 2026-01-17
Support FOSS tests running locally


nettle:
### Describe the task

We need to support multiple bazel versions (6, 7, 8)
when running FOSS tests locally, not only from CI 🤦
As of now we copy just .bazelversion with fixed old version.

### Scope

1. Remove .bazelversion from templates
   See test/foss/templates/.bazelversion
2. Use .bazelversion only from root directory in CI
3. Change CI to create .bazelversion in root directory (the same scenario for ALL tests!)
4. Add more logging
5. Fix problem that **foss/*/test-proj** directories cannot be removed 🐞
6. Add possibility (script?) to run FOSS tests locally
7. Make sure FOSS tests work both in CI and locally the same way!

### Definition of Done

Can successfully run and debug FOSS tests locally

nettle:
OMG! Just found out that FOSS tests depend on CI where it copies `.bazelversion` file.
So, no chance to test locally?

PR #149, opened 2026-01-17
Support FOSS tests running locally

nettle:
Why:
We need to support multiple bazel versions (6, 7, 8)
when running FOSS tests locally, not only from CI.
As of now we copy just .bazelversion with fixed old version.

What:
- Remove .bazelversion from templates
- Copy .bazelversion from root directory
- Change CI to create .bazelversion in root directory
- Print out versions in CI for FOSS tests
- Add more logging (seen only with -vvv)
- Fix small logging bug in test_external_repo.py

Addresses:
#148


test/unit/external_repository/test_external_repo.py
```diff
@@ -53,7 +53,7 @@ def setUpClass(cls):
             logging.debug("No bazel version set, using system default")
         _, stdout, _ = cls.run_command("bazel --version")
         cls.BAZEL_VERSION = stdout.split(' ')[2].strip()
-        logging.debug("Using Bazel", cls.BAZEL_VERSION)
+        logging.debug("Using Bazel %s", cls.BAZEL_VERSION)
```
Szelethus: Nice catch.

Szelethus (commented):
I quite like the change! @furtib? You are the expert on this one.

Szelethus (changes requested):
On a second thought: The patch title is very different to what the patch does. Have we ever explicitly **not** supported running foss tests locally? Is this patch fixing an issue with that? This feels like a CI refactor more than anything.

furtib (changes requested):
The title of the PR does not match what it's doing.
This PR makes no changes to how FOSS tests are run locally, but how they're run in CI.
Otherwise, I see no problem with the patch.

Szelethus:
Let's abandon this PR in favor of the superseding PR.

PR #162, opened 2026-01-21
Bazel 9

furtib:
Note:
codechecker_test works, per_file fails (I think Popen must be installed through rules_python-s pip).
I wrote this without any regard to backward compatibility.
Why:
It was fun

What:
- basically just a bunch of load statements

Addresses:
#108


Szelethus (commented):
Oh, what a joyous time to be alive, even more backward- and forward incompatible bazel versions to support :)

Issue #157, opened 2026-01-20
Add more tests with per_file = False


nettle:
### Describe the task

It is easier to compare new implementation (per_file = True) with reference, old and stable implementation (per_file = False).
Besides old implementation should be tested too.

### Scope

1. Add more test cases with `per_file = False`
2. Make sure all tests pass

### Definition of Done

Ideally all tests should have old and new versions

PR #170, opened 2026-01-27
Add monolithic test for caching

furtib:
Why:
We want comparable testing for all `per_file` test in the original "monolithic" rule.

What:
- Added a test case for (the lack of) catching on the monolithic rule

Addresses:
#157


Szelethus (approved):
LGTM

Issue #153, opened 2026-01-20
PyLint


nettle:
### Describe the task

Introduce `pylint` for python and potentially bazel code verification

### Scope

1. Find an elegant and reliable way to add `pylint` to the project
2. Define very simple, first iteration of `pylint` workflow and rules
3. Suggest simple, easy-to-use `pylint` configuration
4. Plan cleanup and CI integration as next steps

### Definition of Done

- Clear for all developers how to use `pylint` in the project
- Simple requirements are described
- CONTRIBUTING is updated

PR #171, opened 2026-01-28
Add pylint

furtib:
Why:
To enforce coding standards

What:
- Added requirements.txt
- Added a CI workflow for pylint
- Disabled checks:
	- fixme - We use this to indicate a test being disabled
	- in legacy - code-duplication: This test is mostly untouchable; code duplication is unavoidable
- Disabled files:
	- template - Errors in this file are to be fixed by the copier. (e.g. `self.assertTrue(True)` or setUp methods are considered issues by pylint, while they are just unfinished)

Note:
- Developers can easily obtain PyLint using `requirements.txt`.

Addresses:
Fixes: #153

### DIFF
--- a/test/unit/legacy/test_legacy.py
+++ b/test/unit/legacy/test_legacy.py
@@ -81,7 +81,6 @@ class TestBase(unittest.TestCase):
                     logging.debug(line)
                     return line
         self.fail(f"Could not find r'{regex}' in '{filename}'")
-        return ""
 
 class TestBasic(TestBase):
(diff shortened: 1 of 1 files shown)


Szelethus (commented):
I suppose we should fix the issues pylint found, and then merge.

furtib:
Okay, I will remove the Makefile and try to make running the linters easier locally in a follow-up patch.

The argument against changing test.sh is that that specific file is used by a testing CI, which does not have pylint installed.

PR #174, opened 2026-01-29
Add rules_python toolchain to the rule

furtib:
Why:
Necessary for Bazel 8 support.

What:
- Added `rules_python` as a dependency
- Removed registration of our own toolchain from the `MODULE.bazel`; if left there, its priority would shadow the rules_python toolchain. Bazel 7, during unittest, seems to prefer our own, while if included by another project, it seems to prefer the rules_python toolchain. (Adding the following flag will reveal the process of toolchain selection: `--toolchain_resolution_debug=.*`)
(The logic behind this:
	- Bazel 6 prefers the WORKSPACE, so it finds the old toolchain.
	- Bazel 7 uses the MODULE system, but also checks the WORKSPACE file, so it finds the old toolchain too. In projects including it with only a `MODULE` file, e.g., zlib-modules, it will use the rules_python toolchain. (My theory for this is that register_toolchains takes the highest precedence in the order of toolchain resolution, and in other projects, we don't register the toolchain)
	- Bazel 8 ignores the WORKSPACE file, so it does not find the old toolchain; they can't interfere with each other.
)
- Changed from using shebang to changing the executor. This is necessary, since the interpreter provided by rules_python is inside the sandbox, and we will get a relative path, not an absolute one. This is why we can't use shebang. Also, if `use_default_shell_env` is enabled, the path to an executable is accepted as executable.
- Created that absolutely ugly wrapper script for `codechecker_test` - I have no better idea of how I return an executable that, without any parameters, would execute the script using the interpreter, that maybe inside the sandbox, maybe not. I'm open to any better ideas. (The only idea I have is dropping the template expansion entirely, and using py_binary targets instead)
- It is possible to set a specific Python version for the rules_python toolchain. I opted not to do this and let the including project decide which version they want to use.

Addresses:
#108

### DIFF
--- a/src/codechecker.bzl
+++ b/src/codechecker.bzl
@@ -262,10 +262,10 @@ def _codechecker_test_impl(ctx):
         output = ctx.outputs.test_script_wrapper,
         is_executable = True,
         content = """
-            {} {}
+            {python_bin} {test_script}
         """.format(
-            python_interpreter_path,
-            ctx.outputs.codechecker_test_script.short_path,
+            python_bin = python_interpreter_path,
+            test_script = ctx.outputs.codechecker_test_script.short_path,
         ),
     )
 


MODULE.bazel
```diff
@@ -15,10 +15,13 @@ limitations under the License.
...
 
-python_extension = use_extension(
-    "//src:tools.bzl", "module_register_default_python_toolchain")
+bazel_dep(name = "rules_python", version = "0.38.0")
+bazel_dep(name = "rules_cc", version = "0.2.16")
+#python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+#python.toolchain(
+#    python_version = "3.10",
+#    is_default = True,
+#)
```
Szelethus: What is the story with these?

Szelethus (commented):
> Bazel 7, during unittest, seems to prefer our own, while if included by another project, it seems to prefer the rules_python toolchain. 
What does "prefer" mean? Does this mean that we are committing to supporting both WORKSPACE and Module system for bazel 7?

I want to know whether this is an actual requirement, because the way you explain it, this seems to be a non-trivial complexity. Would the code look significantly cleaner if we dropped WORKSPACE support for bazel 7?

src/common.bzl
```diff
@@ -61,12 +61,31 @@ def python_path(ctx):
     elif hasattr(py_toolchain, "py3_runtime"):
         py_runtime = py_toolchain.py3_runtime
         python_path = py_runtime.interpreter_path
+
+        # @rules_python does not provide interpreter_path, but interpreter.path
+        if python_path == None:
+            python_path = py_runtime.interpreter.path
```
Szelethus:
```suggestion
        if hasattr(interpreter_path, python_path):
            python_path = py_runtime.interpreter_path
        elif hasattr(interpreter.path, python_path):
            python_path = py_runtime.interpreter.path
        else:
            fail("Failed to retrieve python path!")
```

src/codechecker.bzl
```diff
@@ -226,27 +229,57 @@ def _codechecker_test_impl(ctx):
...
             "{Mode}": "Test",
             "{Verbosity}": "INFO",
-            "{PythonPath}": python_path(ctx),  # "/usr/bin/env python3",
             "{codechecker_bin}": CODECHECKER_BIN_PATH,
             "{codechecker_files}": codechecker_files.short_path,
             "{Severities}": " ".join(ctx.attr.severities),
         },
     )
 
+    # If we use our toolchain, the path will be an absolute path,
```
Szelethus:
```suggestion
    # If we use our custom toolchain, the path will be an absolute path,
```

src/codechecker.bzl
```diff
@@ -226,27 +229,57 @@ def _codechecker_test_impl(ctx):
...
+            ctx.outputs.codechecker_test_script.short_path,
+        ),
+    )
+
     # Return test script and all required files
-    run_files = default_runfiles + [ctx.outputs.codechecker_test_script]
+    run_files = default_runfiles + [
+        ctx.outputs.codechecker_test_script,
+        ctx.outputs.test_script_wrapper,
+    ] + python_interpreter_tool(ctx)
```
Szelethus: @nettle I understand you find this solution inelegant, and I won't contest it. However, would it be agreeable for you if we pushed this PR as-is, and found a nicer solution after the fact? This code is not the architectural backbone of the patch, so I think it is fine to improve it later.

PR #177, opened 2026-01-30
Fix pylint issues in tests

furtib:
Why:
We want pylint to pass on all python file.

What:
- Applied pylint suggestions

Addresses:
#153

### DIFF
--- a/test/unit/external_repository/test_external_repo.py
+++ b/test/unit/external_repository/test_external_repo.py
@@ -39,7 +39,9 @@ class TestImplDepExternalDep(TestBase):
         Copy bazelversion from main, otherwise bazelisk will download the latest
         bazel version.
         """
-        # The folder bazel-external_repository contains this script
+        # The folder bazel-external_repository
+        # created by bazel during bazel build/test
+        # contains a copy of this test file
         # and the unittest test discovery finds it.
         # This is why, it is imperative that these directories get cleared
         cls.run_command("bazel clean")
@@ -49,6 +51,8 @@ class TestImplDepExternalDep(TestBase):
             shutil.copy(
                 "../../../.bazelversion", "third_party/my_lib/.bazelversion"
             )
+        # If no such file exists assume user doesn't use bazelisk
+        # This file is not needed
         except FileNotFoundError:
             logging.debug("No bazel version set, using system default")
         _, stdout, _ = cls.run_command("bazel --version")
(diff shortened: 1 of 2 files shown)


test/unit/external_repository/test_external_repo.py
```diff
@@ -66,50 +67,50 @@ def tearDownClass(cls):
         cls.run_command("bazel clean")
         try:
             os.remove(".bazelversion")
-        except:
+        # pylint: disable=broad-exception-caught
+        except Exception:
             pass
```
Szelethus: Catching the generic exception and silencing it is a cardinal sin in my view. If the intent is that this file shouldn't exist anyway (which is IIRC the idimatic way of handling this), then we should check against the specific exception type.

Szelethus (changes requested):
_Almost_ LGTM.

test/common/base.py
```diff
@@ -181,7 +185,9 @@ def start_codechecker_server(cls):
             "--port",
             "8001",  # user running unittest must make this port free!
         ]
-        cls.devnull = open(os.devnull, "w")
+        # pylint: disable=consider-using-with
+        cls.devnull = open(os.devnull, "w", encoding='utf-8')
+        # pylint: disable=consider-using-with
         cls.server_process: subprocess.Popen = subprocess.Popen(
             server_command, stdout=cls.devnull
         )
```
Szelethus: I recall in the main codechecker repo, we have had issues with servers not wanting to die. I don't want to pain the devil on the wall, I'm just anxious whether we properly ensure that the server is cleaned up. Are we sure that in the event of an assertion failure or in the event of a crash, the server cleanup runs properly every time?

test/unit/generated_files/test_generated_files.py
```diff
@@ -69,15 +72,17 @@ def test_genrule_source_codechecker(self):
...
-            f"bazel test //test/unit/generated_files:per_file_genrule_header"
+            "bazel test //test/unit/generated_files:per_file_genrule_header"
         )
         self.assertEqual(ret, 3)
         test_log = os.path.join(
-            self.BAZEL_TESTLOGS_DIR, # type: ignore
+            self.BAZEL_TESTLOGS_DIR,  # type: ignore
             "per_file_genrule_header",
-            "test.log"
+            "test.log",
```
Szelethus: Do we need this comma here and in the other function calls?

PR #178, opened 2026-01-30
Apply pylint suggestions to codechecker_script.py

furtib:
Why:
We want to add pylint to CI

What:
- Applied all the pylint suggestions

Note:
Pay attention to this PR! The changes suggested by pylint were quite large.

Addresses:
#153


Szelethus (changes requested):
It feels like whenever the fix from the pylint warning is trivial, we do it, otherwise we just suppress it. That is not the way to interact with static analyzers, not unless each of those suppressions are justified. 

Please reconsider each warning suppression; if they are warranted, write comments that explain it. It should be a last resort solution though.

PR #181, opened 2026-02-02
Provide necessary debug infromation in non-verbose test runs.

furtib:
Why:
I want to be able to get enough information without having to scroll through pages of logs.
By default, pytest hides the current logs. I need to pass --log-cli-level=DEBUG to each run, to see them.

What:
- Add stderr as message for process asserts

Addresses:
none


furtib:
The idea behind #176 is to show the test author that they need to pass stderr as a 3rd parameter.
It doesn't feel like the best solution, but I have no better idea on how to improve it.

Issue #155, opened 2026-01-20
Buildifier


nettle:
### Describe the task

Introduce `buildifier` for bazel code verification

### Scope

1. Find an elegant and reliable way to add `buildifier` to the project
2. Define very simple, first iteration of `buildifier` workflow and rules
3. Suggest simple, easy-to-use `buildifier` configuration
4. Plan cleanup and CI integration as next steps

### Definition of Done

- Clear for all developers how to use `buildifier` in the project
- Simple requirements are described
- CONTRIBUTING is updated

nettle:
I think it would be good to be able to run buildifier locally first :)

PR #182, opened 2026-02-02
Add Buildifier

furtib:
Why:
We want to lint not just our Python but also our `.bzl` files.

What:
- Add a CI job for Buildifier

Addresses:
#155

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -24,15 +24,26 @@ jobs:
 
     - name: Analyzing the code with Buildifier
       run: |
+        set +e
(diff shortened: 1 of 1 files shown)


Szelethus:
I totally agree with the effort, but as it's configured now, this is one of the most horrendous linter outputs I've ever seen. What am I to do with this?:
```
15a16
> 
19c20,23
<     "//src:tools.bzl", "module_register_default_python_toolchain")
---
>     "//src:tools.bzl",
>     "module_register_default_python_toolchain",
> )
> 
```

PR #184, opened 2026-02-02
Optimize CI for billing

furtib:
Why:
We use an unnecessarily large amount of minutes in GitHub Actions.

This happens because billing rounds up to a whole minute, so setup actions (like setup unit/foss) cost a minute instead of 2 seconds. Also, since we run multiple machines, we pay for the environment setup for each (~1 minute for each machine).

Cost in #181: (complete CI run in 6 minutes, cost: 38 minutes, ~8 minutes environment setup)
- 4x6 minutes - FOSS tests
- 4x3 minutes - Unit tests
- 2x1 minutes - Setup jobs

Running sequentially: complete CI: ~8 minutes, cost: 25 minutes (~4 minute environment setup)
- 5 min - unit rhel
- 6 min - unit ubuntu
- 6 min - foss rhel
- 8 min - foss ubuntu

Reorganizing jobs to be run sequentially saves us: ~13 minutes in billed minutes (for every PR), and loses us ~3 minutes in waiting for CI to finish.

Unit tests are usually ~2 minutes long, while FOSS tests are usually ~4 minutes long.
This means that while the current CI setup will stay 6 minutes long, even with Bazel 8, the sequential run will be ~12 minutes in length. (Expected increase in billed minutes: current: +18, optimized: +12, saved: 19 minutes) 

What:
- Removed matrix from CI

Addresses:
none


Szelethus (commented):
We really need to know what our budget is. I'll try to find out ASAP. If it's generous enough, saving those two minutes might be justifiable.

furtib:
Due to the upcoming refactoring of how tests are run and the consequent CI changes, this patch is irrelevant.

Also, we did not run out of our budget.

Issue #163, opened 2026-01-21
Consider tools for multi-configuration testing


nettle:
### Describe the task

We need to support multi-configuration testing
- Bazel versions 6, 7 and 8
- CodeChecker versions 6.25, 6.26, 6.27
- probably different Python versions
- might be different Clang and Clang-Tidy versions
- It would be also good to install and use dev tools like `pylint`, `pycodestyle`, `buildifier`

### Scope

1. Define requirements for multi-configuration testing
2. Consider different tools
3. Discuss pros/cons, benefit and limitations
5. Suggests ways forward: tools, approaches
6. Agree on tools, approaches

### Definition of Done

Discussed and agreed on tools and approaches for multi-configuration testing

nettle:
Meanwhile I'm thinking about something like:
```
├── .ci
│   ├── mise
│   ├── micromamba
│   ├── jenkins
│   └── something-else
├── .github
│   └── workflows
├── src
└── test
```
What do you think?

furtib:
I like the directory structure; it makes things much cleaner.
I wish .github could be placed under `.ci` too.

I actually tried Nox myself, just opened a PR with it (#185, this really is just a show it can do it PR, I haven't thought about details that much). I found it very customizable and easy to work with.

Under license troubles, I meant the Anaconda TOS. I would link it here, but I can't even open their website, hence my lack of knowledge about the issue at hand.
My understanding is that as long as we use the conda-forge repository, we are in the clear.

PR #185, opened 2026-02-03
Showcase Nox with micromamba

furtib:
Why:
We want a stable testing framework that supports testing with different Bazel, Python, and Codechecker versions

What:
- Added a minimal Nox example for testing

Notes:
- The runtime of the full tests is expected to be ~160 minutes.
- This is currently just a showcase of how it could work. I haven't put much thought into details, like placement of files, file names, etc.

Addresses:
#163


.github/workflows/nox.yaml
```diff
@@ -0,0 +1,53 @@
...
+    strategy:
+      fail-fast: false
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          pip install nox pytest
```
nettle: The main point of micromamba is to provide a way to install all dependencies.
If we need to install nox and pytest before micromamba then why do we need micromamba?

furtib: Yeah, I didn't consider that part.

PR #187, opened 2026-02-03
Add Buildifier

furtib:
Why:
 We want to lint not just our Python but also our `.bzl` files.

What:
- Add buildifier_prebuilt as a bazel dependency
- Create test rule for buildifier_prebuilt
- ordered dictionaries
- fixed load statementes missing
Addresses:
#155


furtib:
Sorry, I messed up the previous PR.
We were in a state where the output was illegible, and I tried a couple of things to improve readability.

I tried my best to find a more legible output format.
The best so far is instead of checking the logs, looking at the [summary tab](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#adding-a-job-summary) under the action. (I somehow even got syntax highlighting to work there, at the cost of having to run buildifier a third time, the first 2 are unavoidable)

Szelethus (approved):
LGTM!

BUILD
```diff
@@ -0,0 +1,31 @@
+load("@aspect_rules_lint//format:defs.bzl", "format_test")
+load("@buildifier_prebuilt//:rules.bzl", "buildifier_test")
+
+buildifier_test(
+    name = "buildifier_native",
```
nettle, line 5: I wonder why it is named "_native"?
Any reasons for that?

MODULE.bazel
```diff
@@ -19,6 +19,13 @@ module(name = "rules_codechecker")
 bazel_dep(name = "rules_python", version = "0.32.0")
 bazel_dep(name = "rules_cc", version = "0.2.3")
 
+bazel_dep(
+    name = "buildifier_prebuilt",
+    version = "6.4.0",
+    dev_dependency = True,
+)
+bazel_dep(name = "aspect_rules_lint", version = "1.11.0", dev_dependency = True)
```
nettle, line 27: I think we do not need this. This is actually duplication of buildifier_prebuilt

BUILD
```diff
@@ -0,0 +1,31 @@
...
+    starlark = "@buildifier_prebuilt//:buildifier",
+    starlark_check_args = [
+        "-lint=warn",
+        "-warnings=all",
+        "-mode=diff",
+        # -u will always get passed to buildifier not diff_command
+        #"-diff_command=\"diff -u\"",
+    ],
+    workspace = "//:WORKSPACE",
+)
```
nettle, line 31: Seems like not needed

BUILD
```diff
@@ -0,0 +1,31 @@
+load("@aspect_rules_lint//format:defs.bzl", "format_test")
+load("@buildifier_prebuilt//:rules.bzl", "buildifier_test")
+
+buildifier_test(
+    name = "buildifier_native",
+    diff_command = "diff -u",
```
nettle, line 6: May I ask why do we need this?

Issue #152, opened 2026-01-20
Separate server, parse and store functions from TestBase


nettle:
### Describe the task

Separate server, parse and store functions from TestBase.
Reason: do not mix simple and complex functionality, but separate and extend.

### Scope

1. Add a new unittest base class with server, parse and store functionality testing
2. Use new class only in unittest which requires server, parse, store
3. Make sure nothing is broken

### Definition of Done

- Separate unittest class with server, parse and store functionality
- Nothing is broken

PR #188, opened 2026-02-04
Refactor codechecker_server

### LINKED ISSUE #186: Make CodeChecker server clean up after itslef even if interrupted
### Describe the task

CodeChecker server can currently stay alive as a zombie process, if the tests are interrupted before it can be cleaned up.

### Scope

- Implement CodeChecker server integration in such a way that it gets cleaned up automatically.

### Definition of Done

- CodeChekcer server is cleaned up automatically

furtib:
Why:
The CodeChecker server might stay alive as a zombie process if interrupted.

What:
- Created a CodeCheckerServer class
- Replaced usage of CodeCheckerServer
- Moved helper functions specific to the parse test into it

Addresses:
Fixes: #186
#152

### DIFF
--- a/test/common/codechecker_server.py
+++ b/test/common/codechecker_server.py
@@ -1,0 +1,105 @@
+# Copyright 2023 Ericsson AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


test/common/codechecker_server.py
```diff
@@ -0,0 +1,105 @@
...
+    Cleans up after itself.
+    """
+    def __init__(self, port="8001"):
+        self.running = False
+        self.port = port
+        self.temp_workspace = tempfile.mkdtemp()
+        self.start_codechecker_server()
+
+    def __del__(self):
+        self.stop_codechecker_server()
```
Szelethus, line 76: I admit I haven't spent terribly long with Python, but this might be the first time I've ever seen a destructor in the wild. My understanding is that, because of the garbage collector's unpredictability, this is not considered a Pythonic best practice. Even though your case is fine, as you call it explicitly, that kind of defeats the purpose of this function.

Then again, I'm not terribly opposed to it, because lingering zombie codechecker servers are quite annoying. Considering that we reuse the same server for all tests, this is likely fine.

Szelethus (approved):
LGTM

PR #189, opened 2026-02-04
Refactor cache test to use tempfiles

furtib:
Why:
The current setup might leave artifacts in the folder if interrupted. While these artifacts are automatically deleted at the next test run, this is not an ideal setup.

What:
- replaced the tmp directory with a tempfile object

Addresses:
none


Szelethus (approved):
If this makes your life easier, it shall have my blessings.

PR #196, opened 2026-02-23
Apply buildifier linter suggestions

furtib:
Why:
We want to produce high-quality Bazel code. To do this, we use buildifier to both format and lint our files.

What:
- Applied all\* linter suggestions to bzl files
- Formatted BUILD files with buildifier

\*: In per_file.bzl in the `run_code_checker` function I puposefully ignored the unused variable errors. This function is still in early development, and the unused variables (arguments, label, env_vars, compilation_context) may be necessary in the future.

Addresses:
#155

### DIFF
--- a/src/clang.bzl
+++ b/src/clang.bzl
@@ -215,8 +215,14 @@ def _run_analyzer(
 
 def check_valid_file_type(src):
     """
+    Checks if the file is a cpp related file.
+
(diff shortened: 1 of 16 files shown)


Szelethus (approved):
LGTM!

With regards to `ctx`, yeah, it isn't terribly important, but consistency has its own benefits. I'd keep it.

PR #205, opened 2026-03-16
Show case *most* unittest running through bazel test ...

furtib:
Why:
We want all test to be run through `bazel test ...`

What:
- Created a filegroup in all packages that exports all files inside the package.
- Created a target that runs the Python unittests on the copy inside the sandbox.

Addresses:
none

Note:
This is a proof-of-concept patch.
**The question is: Is this a good way to run tests that cannot be completely integrated into Bazel?**
This solution does not speed up the time it takes to run tests.
It seems like the external test cannot be added in this way. (But it can take FOSS tests **too**

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,0 +1,36 @@
+filegroup(
+    name = "all_sources",
+    srcs = glob(
+        ["**/*"],
+        exclude = [
+            "bazel-*/**",       # Exclude bazel symlinks
+            ".git/**",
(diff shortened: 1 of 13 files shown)


nettle (commented):
That's interesting experiment, but not really a bazel test.
Please go on, but do not submit :)

furtib:
This PR has served its purpose (showcasing that Bazel can be run inside of Bazel)

PR #208, opened 2026-03-17
Refactor FOSS tests

furtib:
Why:
Current FOSS tests are hard to write and have a lot of repeating code.
While the original approach allows highly custom changes to each project, it is too much of a burden to maintain in its current form.

What:
- Rewrote foss tests to use YAML config files (you can comment in these files, unlike in JSON) instead of custom scripts.
- Modified tests to take the Bazel version into consideration during testing.

Addresses:
#148


furtib:
It seems it would be better to do this through bazel.

PR #211, opened 2026-03-31
Add skip option to per_file

### LINKED ISSUE #160: Per-file skipfile option missing
### Describe the task

Currently, the skipfile handling is either missing or broken in the per_file rule.

### Scope

- Implement the skipfile option for the per-file rule

### Definition of Done

Can use a skipfile with the per-file rule.

furtib:
Why:
There is no skipfile option for the per_file rule.
We should strive to solve this on the Bazel side, but in the meanwhile it is okay if we just create an empty plist file for skipped files.

What:
- Added skip option for the per_file rule.

Addresses:
Fixes: #160

### DIFF
--- a/src/per_file_script.py
+++ b/src/per_file_script.py
@@ -18,31 +18,24 @@
 Codechecker wrapper script for per-file analysis
 """
 
-import fnmatch
 import os
-from pathlib import Path
 import re
 import shutil
 import subprocess
 import sys
-from typing import Optional
 
-# The output directory for CodeChecker
-DATA_DIR: Optional[str] = None
-# The file to be analyzed
-FILE_PATH: str = None  # pyright: ignore
-# List of pairs of analyzers and their plist files
-ANALYZER_PLIST_PATHS: list[list[str]] = None  # pyright: ignore
-LOG_FILE: str = None  # pyright: ignore
 COMPILE_COMMANDS_JSON: str = "{compile_commands_json}"
 COMPILE_COMMANDS_ABSOLUTE: str = f"{COMPILE_COMMANDS_JSON}.abs"
 CODECHECKER_ARGS: str = "{codechecker_args}"
 CONFIG_FILE: str = "{config_file}"
-SKIP_LIST: list[str] = ["{skip_list}"]
+SKIP_FILE: str = sys.argv[4]
+# The output directory for CodeChecker
 DATA_DIR = sys.argv[1]
+# The file to be analyzed
 FILE_PATH = sys.argv[2]
 LOG_FILE = sys.argv[3]
-ANALYZER_PLIST_PATHS = [item.split(",") for item in sys.argv[4].split(";")]
+# List of pairs of analyzers and their plist files
+ANALYZER_PLIST_PATHS = [item.split(",") for item in sys.argv[5].split(";")]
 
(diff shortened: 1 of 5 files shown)


Szelethus (commented):
I think the most correct solution would be to make CodeChecker the single source of authority on how to do skipping. For instance, `CodeChecker would-skip mytest.cpp --skip skipfile.txt` could return 0/1. On the other hand... this is quite an established interface we are extremely unlikely to ever change.

@bruntib, can you throw in your two cents?

src/per_file_script.py
```diff
@@ -28,19 +30,55 @@
...
+    <key>diagnostics</key>
+    <array/>
+    <key>files</key>
+    <array/>
+</dict>
+</plist>
+"""
+
+
+def skipped():
```
Szelethus:
```suggestion
def should_skip():
```

Szelethus (commented):
We had an intersting coversation WRT older CC versions. If we patch this CodeChecker-side, it would be more elegant and robust, and wouldn't need much (any?) bazel side changes. However, that would cement skip files not working for per-file analysis with older CC versions. We should consider this carefully.

src/per_file_script.py
furtib: This function has been removed.

src/per_file_script.py
```diff
@@ -118,36 +127,40 @@ def _display_error(ret_code: int) -> None:
...
-                shutil.move(
-                    os.path.join(DATA_DIR, file),   # type: ignore
-                    analyzer_info[1],
-                    )
+    compiled_analyzers = [
+        (re.compile(rf"_{analyzer[0]}_.*\.plist$"), analyzer[1])
+        for analyzer in ANALYZER_PLIST_PATHS
+    ]
+
+    for regex, target_file in compiled_analyzers:
```
Szelethus: Ugh, I see where you are going with this, but this is a little messy. Neither `compiled_analyzers` nor `regex` are terribly talkative. How about leaving the regex compilation in the loop?

src/per_file_script.py
```diff
@@ -118,36 +127,40 @@ def _display_error(ret_code: int) -> None:
...
-    with open(LOG_FILE, "r", encoding="utf-8") as log_file:  # type: ignore
+    with open(LOG_FILE, "r", encoding="utf-8") as log_file:
         print(log_file.read())
     sys.exit(1)
 
 
 def _move_plist_files():
     """
     Move the plist files from the temporary directory to their final destination
+    If the files doesn't exists, write an empty plist file to the target.
```
Szelethus, line 142: Lets rationalize this.
"If the files doesn't exists, write an empty plist file to the target. This can happen when an analysis was skipped because of a given CodeChecker skipfile. Still, for each analysis action we must have an output file, so we substitute it with an empty one.

Szelethus (commented):
LGTM, just please tidy up the code and docs.

src/per_file_script.py
furtib: I would really prefer to leave the compilation out of the loop.
I do understand that regex compilation is still much faster than compilation, but just for readability dropping so much performance...

I tried to make the variables more talkative. Is this better?

Szelethus (approved):
LGTM

PR #212, opened 2026-03-31
Fix CI on bazel 8

furtib:
Why:
CI shouldn't fail
`bazel --version` is missing in bazel 8 on CI
`bazel version` prints to stderr on CI with a huge amount of warnings.

What:
- Use `bazel version --gnu_format` instead of `bazel --version`
- Change the method of getting the Bazel version in external tests to use a regex-based approach for getting the Bazel version, since `bazel version --gnu_format` also reports baselisk version.

Addresses:
none

### DIFF
--- a/test/unit/external_repository/test_external_repo.py
+++ b/test/unit/external_repository/test_external_repo.py
@@ -17,6 +17,7 @@ Test external repositories with codechecker
 """
 import logging
 import os
+import re
 import shutil
 import unittest
 from typing import final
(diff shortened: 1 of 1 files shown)


Szelethus (commented):
I'm not conviced. Is it reasonable to support a bazel binary that is so bare-bones it doesn't even support `--version`? How comes this mostly works?

furtib:
After a bit of digging, it seems like we are using an in-dev Bazel version for Bazel 8 unit tests in CI. (8.6.0- (non-git))
Using any binary provided by Bazel itself (release tab), we will never encounter this issue.
This is a related issue on Bazel's issue board. https://github.com/bazelbuild/bazel/issues/22798

So the question is, should we support custom Bazel forks or not?

test/unit/external_repository/test_external_repo.py
```diff
@@ -55,8 +56,12 @@ def setUpClass(cls):
         # This file is not needed
         except FileNotFoundError:
             logging.debug("No bazel version set, using system default")
-        _, stdout, _ = cls.run_command("bazel --version")
-        cls.BAZEL_VERSION = stdout.split(" ")[2].strip()
+        _, stdout, stderr = cls.run_command("bazel version")
```
nettle: Please use `bazel version --gnu_format` instead

nettle (dismissed):

test/unit/external_repository/test_external_repo.py
Szelethus: Awesome!!!

Szelethus (approved):
LGTM

PR #214, opened 2026-04-07
Downgrade rules_cc in external unittest

furtib:
Why:
The version selected for rules_cc is too new for some bazel versions causing the tests to fail.

What:
- Downgraded rules_cc

Addresses:
none


Szelethus (changes requested):
I understand that this is because bazel 6.2.1 has compatibility issues. My stance is that we should drop pre-bazel6.5.0 support.

Issue #107, opened 2025-11-27
Create a codechecker_store rule.


furtib:
### Describe the task

`codechecker_store` should be a Bazel rule that takes the output artifacts of a CodeChecker analysis and stores them on a remote CodeChecker server.
Preferably, this rule should be run with `bazel run //:target_name`.

The rule should be able to:
- obtain result files of `codechecker_test` targets from remote executors.
- handle multiple `codechecker_test` targets as dependencies.
- handle correcting the paths inside plist files.
- be configurable (url, analysis_name, targets, remote_prefix, credentials)

Proposed attributes:
```
codechecker_store(
    name = "store_analysis",
    tests = [":some_codechecker_test", ":some_other_cc_test"],
    url = "127.0.0.1:8001",
    analysis_name = "my-analysis",
    remote_prefix = "/worker/root/path_to_file",  # optional
    credentials = ":codechecker_credentials", # optional
)
```

### Scope

- Rule obtains analysis result files and stores them on a CodeChecker server.
- Rule fixes paths in plist files obtained from remote executors.

### Definition of Done
1. `codechecker_store` works on locally run analysis, without complex credentials and path fixes.
2. `codechecker_store` obtains plist files from remote executors.
3. `codechecker_store` fixes the paths in the plist files from the remote executors.
4. `codechecker_store` works with remote executors.

PR #216, opened 2026-04-07
Implement store_codechecker rule

furtib:
Why:
Users wanted an easy way to store 

What:
- Added custom rule for storing results of codechecker analysis.
- Added test for the new custom rule.

Addresses:
#107


nettle (changes requested):
As I mentioned before I will reject any changes related to CodeChecker store implementation until the design is presented and approved. Please contact me for details

nettle (changes requested):
Please explore use-cases and existing implementations before suggesting new one.
Let's discuss this offline

PR #217, opened 2026-04-07
Fix buildifier setup

nettle:
Why:
Buildifier setup is a bit excessive.
It can be improved.

What:
- Remove duplicating aspect_rules_lint
- Uplift buildifier version
- Silence warnings from compiler

Addresses: #155

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,31 +1,14 @@
-load("@aspect_rules_lint//format:defs.bzl", "format_test")
 load("@buildifier_prebuilt//:rules.bzl", "buildifier_test")
 
 buildifier_test(
-    name = "buildifier_native",
-    diff_command = "diff -u",
+    name = "buildifier",
     exclude_patterns = [
         "./.git/*",
     ],
     lint_mode = "warn",
+    lint_warnings = ["all"],
     mode = "diff",
     no_sandbox = True,
-    workspace = "//:WORKSPACE",
-)
-
-format_test(
-    name = "format_test",
-    # Temporary workaround for not being able to use -diff_command
-    env = ["BUILDIFIER_DIFF='diff -u'"],
-    no_sandbox = True,
-    # TODO: extend with pylint
-    starlark = "@buildifier_prebuilt//:buildifier",
-    starlark_check_args = [
-        "-lint=warn",
-        "-warnings=all",
-        "-mode=diff",
-        # -u will always get passed to buildifier not diff_command
-        #"-diff_command=\"diff -u\"",
-    ],
+    verbose = True,
     workspace = "//:WORKSPACE",
 )
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -21,10 +21,9 @@ bazel_dep(name = "rules_cc", version = "0.2.3")
 
 bazel_dep(
     name = "buildifier_prebuilt",
-    version = "6.4.0",
+    version = "7.3.1",
     dev_dependency = True,
 )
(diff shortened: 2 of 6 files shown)


BUILD
```diff
@@ -1,31 +1,14 @@
-load("@aspect_rules_lint//format:defs.bzl", "format_test")
 load("@buildifier_prebuilt//:rules.bzl", "buildifier_test")
 
 buildifier_test(
-    name = "buildifier_native",
-    diff_command = "diff -u",
```
furtib, line 6: I strongly prefer `diff -u`, it's much more legible than the default.

MODULE.bazel
```diff
@@ -21,10 +21,9 @@ bazel_dep(name = "rules_cc", version = "0.2.3")
 
 bazel_dep(
     name = "buildifier_prebuilt",
-    version = "6.4.0",
+    version = "7.3.1",
```
furtib, line 24: Correct me if I'm wrong, but Buildifier versions are matching their Bazel "counterpart".

This is why I choose 6.4 to match the lowest version we still support.

BUILD
```diff
@@ -1,31 +1,14 @@
...
-    starlark_check_args = [
-        "-lint=warn",
-        "-warnings=all",
-        "-mode=diff",
-        # -u will always get passed to buildifier not diff_command
-        #"-diff_command=\"diff -u\"",
-    ],
+    verbose = True,
     workspace = "//:WORKSPACE",
 )
```
furtib, line 14: The rule provided by rules_lint is more versatile.

While after further research, I came to the conclusion that I won't be able to extend it with pylint, this target can still be extended with a Python formatter (e.g. Ruff).

I'm not absolutely against removing this, but please do consider the expandability aspect.

furtib (commented):
I agree that I did rush the buildifier patch a bit, and we would benefit from clearing it up.

However, I have some concerns about this change, see my comments.

MODULE.bazel
nettle: I see. I'm actually not sure there are matching, but you might be right

BUILD
nettle: I've just run with and without, and I see no difference

BUILD
nettle: Yes, I do consider possible extensibility of aspect_rules_lint,
however, seems like it is not providing any useful features.
Besides, aspect_rules_lint just uses other rules just wrapping them,
so I could not find why we might need that wrapper.

furtib: Okay, this is fair.

I also noticed it was creating problems in older Bazel 7 versions. (e.g. 7.3.1) so we should get rid of it for that.

BUILD
furtib: Huh, it does seem so.
I wonder why I got a different output earlier.
Previously, I got output like:
```
BUILD:
12d11
<     workspace = "//:WORKSPACE",
13a13
>     workspace = "//:WORKSPACE",
exit status 1
```

furtib (changes requested):
Please also remove:
https://github.com/Ericsson/rules_codechecker/blob/3c619271f2a2e7a01165b18b0df00db72519d884/WORKSPACE#L54-L78

Otherwise LGTM!

furtib (approved):

furtib (changes requested):
Please remove:
https://github.com/Ericsson/rules_codechecker/blob/3c619271f2a2e7a01165b18b0df00db72519d884/WORKSPACE#L54-L78

PR #218, opened 2026-04-08
Define how to run tests

furtib:
Why:
The requirements for running tests are not well defined.
This patch aims to define which targets should be run for testing.
This patch does not define how the user should set up their environment, only how the tests should be run. That is a todo for the next patch.

What:
- Added a script that runs all checks.
-  Modified the `CONTRIBUTING.md` to specify which tests must be run

Addresses:
none?

### DIFF
--- a/.ci/micromamba/init.sh
+++ b/.ci/micromamba/init.sh
@@ -69,10 +69,10 @@ if [[ "$VERBOSE" -ge 2 ]]; then
     micromamba info
 fi
 
-# Clean up previous environment
-info "Cleaning up previous environment [$ENV_NAME]..."
-chmod -R +w $THIS_DIR/micromamba/envs/dev
-rm -rf $THIS_DIR/micromamba/envs/dev
+# Make previous environment modifiable to avoid bazel cache related micromamba errors
+info "Changing permissions for previous environment [$ENV_NAME]..."
+chmod -R +w $THIS_DIR/micromamba/envs/$ENV_NAME
+
 micromamba deactivate
 # Create environment
 info "Creating environment [$ENV_NAME]..."
--- a/.ci/micromamba/init.sh
+++ b/.ci/micromamba/init.sh
@@ -69,7 +69,7 @@ if [[ "$VERBOSE" -ge 2 ]]; then
     micromamba info
 fi
 
-# Make previous environment modifiable to avoid bazel cache related micromamba errors
+# Make the leftover environment modifiable to allow micromamba to delete bazel cache folders during its setup
 info "Changing permissions for previous environment [$ENV_NAME]..."
 chmod -R +w $THIS_DIR/micromamba/envs/$ENV_NAME
 
--- a/.ci/micromamba/init.sh
+++ b/.ci/micromamba/init.sh
@@ -69,9 +69,13 @@ if [[ "$VERBOSE" -ge 2 ]]; then
     micromamba info
 fi
 
+# Clean up previous environment
+info "Cleaning up previous environment [$ENV_NAME]..."
+chmod -R +w $THIS_DIR/micromamba/envs/dev
+rm -rf $THIS_DIR/micromamba/envs/dev
+micromamba deactivate
 # Create environment
 info "Creating environment [$ENV_NAME]..."
-micromamba deactivate
 micromamba create --file $THIS_DIR/$ENV_NAME.yaml --name $ENV_NAME --yes $MAMBA_VERBOSITY
 # micromamba config set env_prompt "[{name}] "
 micromamba activate $ENV_NAME
--- a/.ci/micromamba/run_tests.sh
+++ b/.ci/micromamba/run_tests.sh
@@ -1,6 +1,12 @@
 #!/usr/bin/env bash
+SCRIPT_DIR=$(dirname "$(realpath "$0")")
+cd $SCRIPT_DIR
+
 echo "Initialize micromamba environment..."
-source ./.ci/micromamba/init.sh
+source ./init.sh
+
+# Change directory to project root 
+cd ../../
 
 echo "Running pylint..."
 pylint .
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,4 +49,15 @@ test/test.sh                   | Functional and unit test runner
 Submitting a patch and testing
 -------
 
-Before submitting any changes please make sure all tests and checks are passed, pylint doesn't show warnings on new code, and fill out the Pull Request template. If you are a new contributor, and the template is confusing, feel free to submit the PR and we will help you iron it out! On how to run or add a new test, please see [test/README.md](test/README.md).
+Before submitting any changes please make sure all tests and checks are passed.
+
+You must run the following tests locally:
+* `pylint .`
+* `bazel test //...`
+* `pytest test`
+
+
+On how to run or add a new test, please see [test/README.md](test/README.md).
+
+
+Finally, fill out the Pull Request template. If you are a new contributor, and the template is confusing, feel free to submit the PR and we will help you iron it out!
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,13 +49,16 @@ test/test.sh                   | Functional and unit test runner
 Submitting a patch and testing
 -------
 
-Before submitting any changes please make sure all tests and checks are passed.
+Before submitting any changes please make sure all tests and checks have passed inside the micromamba environment.
 
+To enable the Micromamba testing environment run `source .ci/micromamba/init.sh`.
+For more information see `.ci/micromamba/README.md`.
 You must run the following tests locally:
 * `pylint .`
 * `bazel test //...`
 * `pytest test`
 
+To run all these test inside the micromamba environment automatically, run `.ci/micromamba/run_tests.sh`.
 
 On how to run or add a new test, please see [test/README.md](test/README.md).
 
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+# This script currently assumes that the user have set up their environment.
+# This means users using bazelisk have set their bazel version.
+# Either with the environment variable: USE_BAZEL_VERSION
+# or with the .bazelversion file.
+pylint .
+bazel test //...
+pytest test/
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,8 +1,4 @@
 #!/usr/bin/env bash
-echo "Clean up previous micromamba environment"
-chmod -R +w .ci/micromamba/micromamba/envs/dev
-rm -rf .ci/micromamba/micromamba/envs/dev
-
 echo "Initialize micromamba environment..."
 source ./.ci/micromamba/init.sh
 
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,16 +1,17 @@
 #!/usr/bin/env bash
+echo "Clean up previous micromamba environment"
+chmod -R +w .ci/micromamba/micromamba/envs/dev
+rm -rf .ci/micromamba/micromamba/envs/dev
 
-cleanup() {
-    # Very important
-    # reactivation of the mamba environment is not possible if skipped
-    bazel clean
-    micromamba deactivate
-}
-
-trap cleanup EXIT
-
+echo "Initialize micromamba environment..."
 source ./.ci/micromamba/init.sh
 
+echo "Running pylint..."
 pylint .
+echo "Running bazel test //..."
 bazel test //...
+echo "Running pytest ..."
 pytest test
+
+echo "Exiting micromamba..."
+micromamba deactivate
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,8 +1,16 @@
 #!/usr/bin/env bash
-# This script currently assumes that the user have set up their environment.
-# This means users using bazelisk have set their bazel version.
-# Either with the environment variable: USE_BAZEL_VERSION
-# or with the .bazelversion file.
+
+cleanup() {
+    # Very important
+    # reactivation of the mamba environment is not possible if skipped
+    bazel clean
+    micromamba deactivate
+}
+
+trap cleanup EXIT
+
+source ./.ci/micromamba/init.sh
+
 pylint .
 bazel test //...
-pytest test/
+pytest test


CONTRIBUTING.md
```diff
@@ -49,4 +49,15 @@ test/test.sh                   | Functional and unit test runner
 Submitting a patch and testing
 -------
 
-Before submitting any changes please make sure all tests and checks are passed, pylint doesn't show warnings on new code, and fill out the Pull Request template. If you are a new contributor, and the template is confusing, feel free to submit the PR and we will help you iron it out! On how to run or add a new test, please see [test/README.md](test/README.md).
+Before submitting any changes please make sure all tests and checks are passed.
```
nettle: I'm sure, before running anything we must crealy describe how to initialize the environment.
See **Development Environment** above

nettle (commented):
Good start! Thanks @furtib!

run_tests.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# This script currently assumes that the user have set up their environment.
+# This means users using bazelisk have set their bazel version.
+# Either with the environment variable: USE_BAZEL_VERSION
+# or with the .bazelversion file.
```
nettle: Hm... discussable...
Maybe we should setup correct environment in this script?

Szelethus: Set up through program arguments/environmental variables? I understand that our clientbase have their own bazel and CodeChecker distributions, so we can't generalize too much here.

This sounds like a reasonable compromise:
`USE_BAZEL_VERSION=8.5.0 CODECHECKER_BIN_PATH=/path/to/codechecker/25 run_tests.sh`

nettle: This is about **Development Environment** not Execution Environment since we run tests, right?
Here we have to find common ground, common environment that works for all possible contributors, starting from you and me, but extend to anyone. Which already means at least different versions of Ubuntu and RedHat, different version sets of Bazel and Clang etc.
That's exactly why I suggested **Mise** and/or **Micromamba** to solve this problem.
Both are still not ideal, but at least something. Maybe some alternatives are possible too.
(BTW, native bazel dependency manager may work, but we did not try)

furtib: We could very well set up the python virtual environment for testing in this script, but I would be hesitant to also do that for the llvm toolchain (maybe thats where im wrong?).

My first idea is whether a bazel provided llvm toolchain exist and can we use that?

My experience is that most of these crashes should be due to diagtool not being found by clang. (You can check the codechecker log for the failing analyzed file to confirm this, even though the plist files arent getting created the log still does)

nettle: > We could very well set up the python virtual environment for testing in this script, but I would be hesitant to also do that for the llvm toolchain (maybe thats where im wrong?).

Right, unfortunately, **venv** (virtual environment) is not sufficient, that's exactly why I suggested **mise** and/or **micromamba**.

> My first idea is whether a bazel provided llvm toolchain exist and can we use that?

Yes, it is theoretically possible!
Practically... I dont know, it might be difficult to find llvm package that we could use. 

> My experience is that most of these crashes should be due to diagtool not being found by clang. (You can check the codechecker log for the failing analyzed file to confirm this, even though the plist files arent getting created the log still does)

There are so many different failures depending on versions and environments, so did not check them.
We need to fix the environment first, then set of tests and the way we run them.
After that I can report and investigate fails.

furtib:
I have investigated the possibility of using the bazel module: https://github.com/bazel-contrib/toolchains_llvm
And I found it to be surprisingly doable. I don't even think the big download size is a huge blocker, since it won't redownload it even after a `bazel clean`.
With that said, it introduces the complexity of whether we require all users to use it or not. (How do we test both methods if we do not require it?)

So instead, I tried this different approach, setting up the environment "manually".
I try not to set any dependency that we need to vary, like Bazel; I'm still figuring out how to handle that.
I chose this because it produces an environment that I think is more easily reproducible to end users compared to what Micromamba provides.

nettle (commented):
This looks to me as introduction of venv + llvm environment :)
It should go under .ci/ as yet another dev env along with mise and micromamba.
And it should be a separate review, not a part of "Define how to run tests".

furtib:
Fair!
I will strip this back to just the script.

That raises some questions about whether the file `test/test.sh` is still being used in CI workflows, or if I can remove that entirely.

nettle:
> Fair! I will strip this back to just the script.

Actually I think venv or uv could be useful :)
But again as separate PRs

> That raises some questions about whether the file `test/test.sh` is still being used in CI workflows, or if I can remove that entirely.

To be honest just adding such script does not bring much value without defining the environment I think.
Without defining reliable environment (micromamba, mise, venv, uv or something else)
we cannot say we have answer to the question "how to run tests?"

nettle:
> I have investigated the possibility of using the bazel module: https://github.com/bazel-contrib/toolchains_llvm And I found it to be surprisingly doable. I don't even think the big download size is a huge blocker, since it won't redownload it even after a `bazel clean`. With that said, it introduces the complexity of whether we require all users to use it or not. (How do we test both methods if we do not require it?)

This is actually interesting!
Can you please do a prototype? (as a separate PR)

> So instead, I tried this different approach, setting up the environment "manually". I try not to set any dependency that we need to vary, like Bazel; I'm still figuring out how to handle that. I chose this because it produces an environment that I think is more easily reproducible to end users compared to what Micromamba provides.

Well, setting up the environment "manually" sounds extremely unreliable.
To make sure we have "common grounds" for tests the setup must be automated.
But of course it starts from describing "manual" steps and conditions.

Szelethus:
I understand that this patch is now beyond a simple "single point of entry" test script.

Lets take a step back, I fear some misunderstandings have to be beaten into my head. The end goal (browsing through our last meeting notes) is the following. We should consider a PR ready to merge (after reviewer approval(s)) if:

* Tests pass locally
* Github CI tests pass
* Jenkins job passes

And of course it would be great if those CI environments could be replicated and run locally with relative ease, with mise/micromamba. What can we do today that isn't a hack and gets us further towards this end goal? I suggest that

1. First, lets finish the discussion on #219 and land it. This PR is somewhat, but not entirely orthogonal to that.
2. We all agree that we should define the set of commands that needs to be run on each patch. We also seem to mostly agree what those commands should be (currently in this PR's run_tests.sh). Lets create a PR that either lands that script bare-bones without any environmental setup, or simply lists those commands in CONTRIBUTING.
3. Proceed with the discussion on the definition of testing environment.

Let's land and finish up the few, small steps we agreed upon, so we don't have so many threads open.

nettle:
> I understand that this patch is now beyond a simple "single point of entry" test script.

It was never been "a simple single point of entry test script" to me :)
The intention is to define how exactly we set the environment (environment manager, tools, versions etc) so that all tests pass always?

> Lets take a step back, I fear some misunderstandings have to be beaten into my head. The end goal (browsing through our last meeting notes) is the following. We should consider a PR ready to merge (after reviewer approval(s)) if:
> 
> * Tests pass locally
> * Github CI tests pass
> * Jenkins job passes
> 
> And of course it would be great if those CI environments could be replicated and run locally with relative ease, with mise/micromamba.

Exactly! That's the goal of this task - how to setup the environment? what tests to run? how?
And all of these must be the same for any local environment and CI.
The results must be repeatable always!

> What can we do today that isn't a hack and gets us further towards this end goal?
> I suggest that
> 
> 1. First, lets finish the discussion on [Recommended requirements #219](https://github.com/Ericsson/rules_codechecker/pull/219) and land it. This PR is somewhat, but not entirely orthogonal to that.

Yes, that we should do. However, #219 is not a showstopper for this task.

> 2. We all agree that we should define the set of commands that needs to be run on each patch. We also seem to mostly agree what those commands should be (currently in this PR's run_tests.sh). Lets create a PR that either lands that script bare-bones without any environmental setup, or simply lists those commands in CONTRIBUTING.

No, this does not help much without setting up reliable environment.
To me such scripts (run_tests.sh and similar) are completely useless if they fail when running in unpredictable environment.

> 3. Proceed with the discussion on the definition of testing environment.

I'm sure this is the main thing we need to do here.
We have some options - micromamba, mise, venv? uv? docker?
Unfortunately, we cannot "equalize" the environment for "standard" OSs like Ubuntu and RedHat, we discussed that.
Unfortunately, docker is also unavailable in E2C/VDI RHEL.
So... then, micromamba, mise, venv? uv? something else?

> Let's land and finish up the few, small steps we agreed upon, so we don't have so many threads open.

I would suggest addressing the main problem - reliable environment.
That's the only serious problem here :)

Szelethus:
We chatted with @furtib a bunch, and I think it'd be best to pursue micromamba! We'll come back with some more concrete ideas once time allows.

run_tests.sh
```diff
@@ -0,0 +1,16 @@
...
+cleanup() {
+    # Very important
+    # reactivation of the mamba environment is not possible if skipped
+    bazel clean
+    micromamba deactivate
+}
+
+trap cleanup EXIT
+
+source ./.ci/micromamba/init.sh
```
nettle: I had previous micromamba dev env in .ci/micromamba/micromamba
so init immediately failed with:

    critical libmamba Non-conda folder exists at prefix - aborting.

I had to remove it completely to unlock.
So... we should work on how to make it robust

run_tests.sh
```diff
@@ -0,0 +1,16 @@
...
+
+pylint .
+bazel test //...
+pytest test
```
nettle, line 16: These steps need some log/diagnostic messages to just indicate what is going on on the screen. Something like:

    echo "Running pylint"
    pylint .
    ...

run_tests.sh
furtib: So just to be sure:
You had a previous micromamba environment, which you used without the script, then when running the script the first time you got this error.

This issue did not return when using the script multiple times, right?

In that case, I believe this error was the exact reason for adding the cleanup trap in the script.
I know that this solution is not ideal, in that sense, micromamba used outside the script (exited without running `bazel clean`) will cause this error to emerge, but I'm not sure how I could force users to run bazel clean before running `micromamba deactivate` (or killing their terminal).

I'm open to any idea on how to make it even just a bit more robust.

furtib: I spent a lot of time seeing how I could make it even just a bit better, and came to the conclusion that deleting the environment is the best way to ensure it will work.

Maybe this cleanup is better placed in the init.sh file, so users are able to use the micromamba environment manually too.

This does not force micromamba to redownload all packages, those are cached under `micromamba/pkgs`.

run_tests.sh
furtib: Fair point.
Done.

.ci/micromamba/init.sh
```diff
@@ -69,9 +69,13 @@ if [[ "$VERBOSE" -ge 2 ]]; then
     micromamba info
 fi
 
+# Clean up previous environment
+info "Cleaning up previous environment [$ENV_NAME]..."
+chmod -R +w $THIS_DIR/micromamba/envs/dev
```
nettle: I think `dev` is actually `$ENV_NAME`

run_tests.sh
nettle: Yes, maybe you are right. I also had to remove previous environment.

.ci/micromamba/run_tests.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
nettle, line 1: What I think is that `run_tests.sh` is actually micromamba dependent, so maybe we should keep it as `.ci/micromamba/run_tests.sh` ?

.ci/micromamba/init.sh
```diff
@@ -69,9 +69,13 @@ if [[ "$VERBOSE" -ge 2 ]]; then
...
+rm -rf $THIS_DIR/micromamba/envs/dev
```
nettle: Hm... Are you sure this is safe? There reason might be running session. Maybe proper deactivation may help?

run_tests.sh
Szelethus: Resolving this conversation as we went, we took the micromamba route.

.ci/micromamba/init.sh
furtib: Good point, thanks!

.ci/micromamba/run_tests.sh
furtib: Okay, I will move it there!

.ci/micromamba/init.sh
furtib: I messed around a bit more; it seems changing the permissions is sufficient.
With this, we make Micromamba take care of the cleanup.

.ci/micromamba/init.sh
```diff
@@ -69,9 +69,13 @@ if [[ "$VERBOSE" -ge 2 ]]; then
     micromamba info
 fi
 
+# Make previous environment modifiable to avoid bazel cache related micromamba errors
```
Szelethus:
```suggestion
# Make the leftover environment modifiable to allow micromamba to delete bazel cache folders during its setup
```

Szelethus (approved):
Make sure to adjust CONTRIBUTING. Otherwise LGTM

.ci/micromamba/init.sh
```diff
@@ -69,9 +69,13 @@ if [[ "$VERBOSE" -ge 2 ]]; then
     micromamba info
 fi
 
+# Make the leftover environment modifiable to allow micromamba to delete bazel cache folders during its setup
+info "Changing permissions for previous environment [$ENV_NAME]..."
+chmod -R +w $THIS_DIR/micromamba/envs/$ENV_NAME
```
Szelethus, line 74: If the folder didn't already exist, I get errors:
```
[micromamba] Changing permissions for previous environment [dev]...
chmod: cannot access '/path/to/repo/.ci/micromamba/micromamba/envs/dev': No such file or directory
```

This is likely just cosmetic, but inelegant.

CONTRIBUTING.md
```diff
@@ -49,4 +49,18 @@ test/test.sh                   | Functional and unit test runner
 Submitting a patch and testing
 -------
 
-Before submitting any changes please make sure all tests and checks are passed, pylint doesn't show warnings on new code, and fill out the Pull Request template. If you are a new contributor, and the template is confusing, feel free to submit the PR and we will help you iron it out! On how to run or add a new test, please see [test/README.md](test/README.md).
+Before submitting any changes please make sure all tests and checks have passed inside the micromamba environment.
+
+To enable the Micromamba testing environment run `source .ci/micromamba/init.sh`.
+For more information see `.ci/micromamba/README.md`.
```
Szelethus, line 55: Btw, we forgot to mention `run_tests.sh` in that file. It would be elegant to either create a link to this documentation file, or just mention the script there as well.

Issue #164, opened 2026-01-21
Execution requirements


nettle:
### Describe the task

Unfortunately, we have very tentatively described prerequisites and missing execution requirements towards environments where the solution might be operating. Operating systems might be different, tools might be different, versions might be different, libraries might be different, settings might be different, and so on and so forth.
So, we should reveal as much as possible dependencies and conditions, limitations and version ranges, and formulate execution requirements,
at least the following:
- Supported operating systems
- Supported versions of Python, Clang, Clang-Tidy, other Clang tools, CodeChecker, Bazel
- Conditions for Clang execution (e.g. CCache, dialtool etc)
- Conditions for virtual environments (e.g. venv, conda, uv, mise etc)
- Describe all revealed conditions and limitations
- Update (if needed) installation steps for typical environments (e.g. Ubuntu, RHEL etc)

### Scope

- Check and define supported operating systems
- Check and define supported versions of Python, Clang, Clang-Tidy, other Clang tools, CodeChecker, Bazel
- Reveal and describe conditions for Clang execution (e.g. CCache, dialtool etc)
- Reveal and describe conditions for virtual environments (e.g. venv, conda, uv, mise etc)
- Describe all revealed conditions and limitations
- Update (if needed) installation steps for typical environments (e.g. Ubuntu, RHEL etc)

### Definition of Done

Documentation is updated to reflect:
- Supported operating systems
- Supported versions of Python, Clang, Clang-Tidy, other Clang tools, CodeChecker, Bazel
- Conditions for Clang execution (e.g. CCache, dialtool etc)
- Conditions for virtual environments (e.g. venv, conda, uv, mise etc)
- Describe all revealed conditions and limitations
- Update (if needed) installation steps for typical environments (e.g. Ubuntu, RHEL etc)

nettle:
Here are currently used versions in 3 actual projects:

Bazel | CodeChecker | OS | clang / clang-tidy | python
--- | --- | --- | --- | ---
7.6.1 | 6.27.3 | RHEL 8.10 & RHEL 9.0 | 23.0 and 19.1.7 | 3.8.16
7.6.1 |  6.27.1 |  RHEL 8.10 & RHEL 9.0 | 23.0 | 3.8.16
7.5.0 | 6.27.3 | Ubuntu | | 3.12

I suggest that for today we define Recommended* requirements as:
Bazel | CodeChecker | OS | clang / clang-tidy | python
--- | --- | --- | --- | ---
[...]

PR #219, opened 2026-04-08
Recommended requirements

nettle:
Why:
We need to agree on recommended requirements
to use for development and run all tests

What:
- Update Bazel and CodeChecker versions
- Update README 

Addresses: #164

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,20 +51,20 @@ Prerequisites
 
 We need the following tools:
 
-- Git 2 or newer (we use 2.36)
-- Bazel 6, 7 or 8 (we recommend version 8.5.0)
-- Clang 16 or newer (we use 16), we use clang-tidy
-- Python 3.8 or newer (we use 3.11)
-- CodeChecker 6.26 or newer (we use 6.26.0)
+- Git 2 or newer
+- Bazel 7.7
+- Clang and clang-tidy 21
+- Python 3.11 or newer
+- CodeChecker 6.27.3
 
 If, by chance, Environment Modules (https://modules.sourceforge.net/)
 are available in your system, you can just add the following modules:
 
     module add git
-    module add bazel/6
-    module add clang/16
+    module add bazel/7
+    module add clang/21
     module add python/3.11
-    module add codechecker/6.26
+    module add codechecker/6.27.3
 
 
 Getting Started
--- a/.ci/micromamba/dev.yaml
+++ b/.ci/micromamba/dev.yaml
@@ -7,8 +7,8 @@ dependencies:
   - pylint
   - pycodestyle
   - buildifier
-  - clang
-  - clang-tools
-  - bazel=6
+  - clang=21
+  - clang-tools=21
+  - bazel=7.7.1
   - pip:
-    - codechecker==6.25
+    - codechecker==6.27.3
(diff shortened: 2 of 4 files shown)


BUILD
```diff
@@ -5,6 +5,7 @@ buildifier_test(
     name = "buildifier_native",
     diff_command = "diff -u",
     exclude_patterns = [
+        "./.ci/*",
```
Szelethus, line 8: Great catch.

Szelethus (commented):
The patch is correct. I posted some comments in the associated issue first.

BUILD
nettle: Unfortunately, still does not work :(

README.md
```diff
@@ -51,20 +51,20 @@ Prerequisites
...
+- Python 3.11 or newer
+- CodeChecker 6.27.3
 
 If, by chance, Environment Modules (https://modules.sourceforge.net/)
 are available in your system, you can just add the following modules:
 
     module add git
-    module add bazel/6
-    module add clang/16
+    module add bazel/7
```
nettle, line 64: There is only Bazel 7.3.1 in "my" RHEL8 environment, which does not work :(
I guess I need to remove all of these "module add" things as not actual.

BUILD
nettle: Now addressed in #220

README.md
```diff
@@ -51,20 +51,20 @@ Prerequisites
 
 We need the following tools:
 
-- Git 2 or newer (we use 2.36)
-- Bazel 6, 7 or 8 (we recommend version 8.5.0)
-- Clang 16 or newer (we use 16), we use clang-tidy
-- Python 3.8 or newer (we use 3.11)
-- CodeChecker 6.26 or newer (we use 6.26.0)
+- Git 2 or newer
+- Bazel 7.7
```
Szelethus, line 55: We added bazel 8 support recently. Among other things, we run regular bazel 8 CI jobs for both Ubuntu and Rhel. I suggest we add 8.5.0 or 8.6.0 as well, that is a rather hard requirement from other parties.

Szelethus (commented):
Reading the specific wording of the summary ("recommended" versions), I agree with sentiment, but I'm wondering we should use more permissive language in the README, like "bazel >=7.7 but <9". Or, we could copy your definition of "recommended" from https://github.com/Ericsson/rules_codechecker/issues/164#issuecomment-4205645991.

README.md
nettle: Well, we should make a decision about **Recommended requirements**.
Is it 7.7 or 8.5 or 8.6?

Szelethus: I suggest that it should be 7.7 _and_ 8.6. My understanding is that within a major Bazel release, upgrading to a newer minor release is straightforward. But, with regards to major versions, if "recommended versions" are, in part, according to your comment, those that 

> 1. We set them as default for Development Environment
> 2. All tests must always pass in the environment with Recommended requirements
> 3. We recommend them to the users as absolutely reliable
> 4. We immediately fix all issues if they are reported in Recommended requirements

beside the first point, both bazel 7 and 8 (and the respective latest minor versions) qualify pretty strongly for that, as there are other teams using these rules that use bazel 8 exclusively.

nettle: I'll keep 7.7 so far, as we discussed, but I think soon we should update it to 8.5

Szelethus (approved):
Send it!

PR #220, opened 2026-04-09
Fix buildifier for micromamba

nettle:
Why:
Buildifier, unfortunately does not work under Micromamba environment
because buildifier scans (find) all subdirectories including bazel output base dir
inside micromamba dir and some files cannot be read there, so it causes error.

What:
- Add patches for buildifier to prune excluded locations
- Test for Bazel 6.5 and 7.7

Test:

    source .ci/micromamba/init.sh
    bazel test :buildifier_native

Addresses: #155, #163

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,1 +1,2 @@
+.ci/micromamba
 test/unit/external_repository
--- a/.ci/buildifier/BUILD
+++ b/.ci/buildifier/BUILD
@@ -1,0 +1,4 @@
+exports_files([
+    "prune.patch",
+    "template.patch",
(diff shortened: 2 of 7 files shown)


nettle:
Hi @furtib, @Szelethus,
Please see how patch technique is working in Bazel ;)
Yes, it is official!

Szelethus (approved):
LGTM

So this is the technique to, as I remember, "use, love and preach". I understand this is pretty much our only option here, considering that buildifier is a third-party library, but I wonder whether this is the solution to persue for certain features.

PR #221, opened 2026-05-01
Proof of concept llvm-toolchain

furtib:
Why:
We want Bazel to handle all dependencies, including the LLVM toolchain.

What:
- Added the LLVM toolchain as a dependency
- Modified the monolithic/single job codechecker rule to use the provided binaries by adding the folder containing them to the PATH

Addresses:
none

Note:
Well that broke spectacularly.....
Issues:
- Each project, including the rule set, will be required to also import the LLVM toolchain (the LLVM toolchain does not allow dependencies to download it) -> This breaks Foss tests and external unit tests,
- Couple legacy tests are failing, will look into it when I have time, some fails due to clang taking precedence over gcc, changing the compile_commands.json content
  

"llvm_toolchain_llvm"  contains every binary, not just clang. I might be mistaken in how I import the toolchain, but I believe we need this one for sure.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -40,3 +40,17 @@ codechecker_extension = use_extension(
     "module_register_default_codechecker",
 )
 use_repo(codechecker_extension, "default_codechecker_tools")
+
+# Last version to support bazel 6
+bazel_dep(name = "toolchains_llvm", version = "0.10.3")
+
+# Configure and register the toolchain.
+llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
+llvm.toolchain(
+   llvm_version = "16.0.0",
+)
+
+use_repo(llvm, "llvm_toolchain")
+use_repo(llvm, "llvm_toolchain_llvm")
+
+register_toolchains("@llvm_toolchain//:all")
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -41,8 +41,8 @@ codechecker_extension = use_extension(
 )
 use_repo(codechecker_extension, "default_codechecker_tools")
 
-# Last version to support bazel 6
-bazel_dep(name = "toolchains_llvm", version = "0.10.3")
+
+bazel_dep(name = "toolchains_llvm", version = "1.7.0")
 
 # Configure and register the toolchain.
 llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -47,10 +47,14 @@ bazel_dep(name = "toolchains_llvm", version = "1.7.0")
 # Configure and register the toolchain.
 llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
 llvm.toolchain(
-   llvm_version = "16.0.0",
+    name = "llvm_toolchain",
+    llvm_version = "20.1.4",
+    extra_llvm_distributions = {
+        "LLVM-20.1.4-Linux-ARM64.tar.xz": "4de80a332eecb06bf55097fd3280e1c69ed80f222e5bdd556221a6ceee02721a",
+        "LLVM-20.1.4-Linux-X64.tar.xz": "113b54c397adb2039fa45e38dc8107b9ec5a0baead3a3bac8ccfbb65b2340caa",
+        "LLVM-20.1.4-macOS-ARM64.tar.xz": "debb43b7b364c5cf864260d84ba1b201d49b6460fe84b76eaa65688dfadf19d2",
+        "clang+llvm-20.1.4-x86_64-pc-windows-msvc.tar.xz": "2b12ac1a0689e29a38a7c98c409cbfa83f390aea30c60b7a06e4ed73f82d2457",
+    },
 )
 
-use_repo(llvm, "llvm_toolchain")
 use_repo(llvm, "llvm_toolchain_llvm")
-
-register_toolchains("@llvm_toolchain//:all")
--- a/src/codechecker.bzl
+++ b/src/codechecker.bzl
@@ -116,10 +116,7 @@ def _codechecker_impl(ctx):
         },
     )
 
-    llvm_bin_dir = ""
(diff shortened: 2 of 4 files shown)


MODULE.bazel
```diff
@@ -31,3 +31,17 @@ codechecker_extension = use_extension(
     "module_register_default_codechecker",
 )
 use_repo(codechecker_extension, "default_codechecker_tools")
+
+# Last version to support bazel 6
```
nettle: I think we should not use MODULE.bazel for Bazel 6

src/codechecker.bzl
```diff
@@ -97,6 +99,16 @@ def _codechecker_impl(ctx):
 
     config_file, codechecker_env = get_config_file(ctx)
 
+    cc_toolchain = find_cpp_toolchain(ctx)
+    clang_executable = cc_toolchain.compiler_executable
+    if hasattr(clang_executable, "dirname"):
+        wrapper_bin_dir = clang_executable.dirname
+    else:
+        wrapper_bin_dir = "/".join(clang_executable.split("/")[:-1])
+    real_llvm_bin_dir = wrapper_bin_dir.replace("llvm_toolchain", "llvm_toolchain_llvm")
```
nettle: I dont understand this part yet :)

nettle (commented):
Thanks @furtib! This is really interesting!
So... as you mentioned we cannot use this approach at least explicitly because C/C++ rules are trying to use this LLVM toolchain :)
We should think how to move forward

MODULE.bazel
furtib: True, and since we are planning on dropping Bazel 6 support, this is no longer a concern.

src/codechecker.bzl
furtib: No wonder I for sure made this wrong.
My goal was to have the directory containing the LLVM libraries in PATH.
The documentation for the toolchain did not hold my hand at all on how this should be done.
I'm looking at their tests right now to figure out how they did it.

furtib:
Well, I made some adjustments, mainly to how the toolchain is recognized.

The new idea is to have an optional parameter where users can provide their llvm_toolchain bin directory as input, and add that to PATH. (This is much better than the one I did before, but it does require more setup from users.)

Does this approach seem reasonable?
I do recognize that defining the llvm_toolchain for each target is not the most ergonomic, but it is the most flexible way. We have seen similar methods in aspect lint. This way, we even support custom LLVM rules.

I believe this would also allow users to de-prioritize the LLVM toolchain in the selection order of C++ toolchains(?).
I'm not sure.

```
format_test(
    name = "format_test",
    starlark = "@buildifier_prebuilt//:buildifier", # <- this line
	...
)
```

What still needs to be done:
- For a patch like this to land, we would need to change our testing. (The tests could have their own `MODULE.bazel`, an example is the [llvm_toolchain repo](https://github.com/bazel-contrib/toolchains_llvm)) This is necessary since llvm_toolchain does not allow itself to be a transitive dependency (due to its size). -> This would fix foss tests, and external tests (these are the ones that fail)

furtib:
To make it a bit more serious, I fixed the tests.
It does make the patch very messy :/

The highlights:
- I copied the testing solution of the [llvm_toolchain](https://github.com/bazel-contrib/toolchains_llvm), so:
    - Tests area separate bazel project, with its own `MODULE.bazel`. (I did have to modify every single test to accompany this :( )
- I modified the `codechecker_pass` target in legacy tests to show how users would use the llvm_toolchain. (This is just for demonstration)

nettle:
> The new idea is to have an optional parameter where users can provide their llvm_toolchain bin directory as input, and add that to PATH. (This is much better than the one I did before, but it does require more setup from users.)
> 
> Does this approach seem reasonable? I do recognize that defining the llvm_toolchain for each target is not the most ergonomic, but it is the most flexible way. We have seen similar methods in aspect lint. This way, we even support custom LLVM rules.
> 
> I believe this would also allow users to de-prioritize the LLVM toolchain in the selection order of C++ toolchains(?). I'm not sure.

Yes, this might be used. Basically this way we can provide path to particular binaries.
And I think this should be at least an option.
However, as I can understand, as soon as we use LLVM toolchain the whole package may want to use it for cc_binary, cc_libary etc. In some cases this is not the scenario users may want.

> What still needs to be done:
> 
> * For a patch like this to land, we would need to change our testing. (The tests could have their own `MODULE.bazel`, an example is the [llvm_toolchain repo](https://github.com/bazel-contrib/toolchains_llvm)) This is necessary since llvm_toolchain does not allow itself to be a transitive dependency (due to its size). -> This would fix foss tests, and external tests (these are the ones that fail)

I dont think creating more packages inside packages is a good idea. At least it is highly discouraged in Bazel.
Besides, what is the purpose?
To be honest I did not get "llvm_toolchain does not allow itself to be a transitive dependency" etc.
I think we need a meeting where you explain.

nettle:
> To make it a bit more serious, I fixed the tests. It does make the patch very messy :/

Well, the recent change made it impossible to review :)
I guess we should probably drop this...

> The highlights:
> 
> * I copied the testing solution of the [llvm_toolchain](https://github.com/bazel-contrib/toolchains_llvm), so:
>   
>   * Tests area separate bazel project, with its own `MODULE.bazel`. (I did have to modify every single test to accompany this :( )

We should definitely avoid packages inside packages.
In general the idea of multiple packages is not good and often an indication of bad design.
I saw projects with similar mistakes - later it took much efforts to fix them.

> * I modified the `codechecker_pass` target in legacy tests to show how users would use the llvm_toolchain. (This is just for demonstration)

Yes, I got the idea, Thanks!

furtib:
> Well, the recent change made it impossible to review :)

Sorry about that, I reverted those changes

>  To be honest I did not get "llvm_toolchain does not allow itself to be a transitive dependency" etc.

This is the offending line:
https://github.com/bazel-contrib/toolchains_llvm/blob/e15a648c23a6305c58f2cb54486ce99fe331696e/toolchain/extensions/llvm.bzl#L55

Not sure what an acceptable way to handle this is.
Lets discuss it in details in a meeting.

(I did make some miscellaneous changes, compared the issue of transitive dependencies it is not important)

PR #222, opened 2026-05-04
Fix README venv instructions

Szelethus:
We should source ./codechecker_venv/bin/activate.

Why:

`./codechecker_venv/` is a directory, it can't be sourced.

What:

Instead, source what we really should, the `acticvate` script.

Addresses:

-


furtib:
LGTM

furtib (approved):
Two times the charm.
LGTM

PR #226, opened 2026-05-19
Introduce CodeChecker toolchain

furtib:
Why:
We want to use toolchains to determine which CodeChecker binary to use.

What:
- Created a toolchain for CodeChecker following the Bazel documentation (https://bazel.build/extending/toolchains)
- Made both CodeChecker rules use the toolchain
- Added documentation for the toolchain

Addresses:
none

Notes:
- Looking around in other toolchains: [rules_go](https://github.com/bazel-contrib/rules_go) and [rules_python](https://github.com/bazel-contrib/rules_python) it seems toolchains are usually under a folder named the program - in out case CodeChecker (python for rules_python, go for rules_go) - this seems to be to have the toolchain path be `//codechecker:toolchain_type` instead of `//src:toolchain_type`.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -38,3 +38,7 @@ codechecker_extension = use_extension(
     "module_register_default_codechecker_tools",
 )
 use_repo(codechecker_extension, "default_codechecker_tools")
+
+register_toolchains(
+    "//src:codechecker_local_toolchain",
+)
--- a/README.md
+++ b/README.md
@@ -427,3 +427,83 @@ After that you can find all artifacts in `bazel-bin` directory:
     
     # compile_commands.json for compile_commands_pass
     cat bazel-bin/test/compile_commands_pass/compile_commands.json
+
+Toolchains
+----------
+
+The previous rules locate the CodeChecker, Clang and clang-tidy executables through a
+Bazel [toolchain](https://bazel.build/extending/toolchains). Modeling the tools
+as a toolchain means the analysis rule (`codechecker_test`) never hardcode binary paths:
+Bazel resolves the correct tools at build time, and you can override them
+per-project or per-platform without touching the rules themselves.
+
+### The default toolchain
+
+For most users no setup is required. `rules_codechecker` ships with a module
+extension that provisions a default set of tools and a pre-registered
+toolchain.
+
+> [!NOTE]
+> The default tools resolve to the `CodeChecker`, `clang` and `clang-tidy`
+> binaries available from PATH on your system (see the [Prerequisites](#prerequisites)).
+
+To enable it, add the following to your `MODULE.bazel`:
+
+```python
+codechecker_extension = use_extension(
+    "@rules_codechecker//src:tools.bzl",
+    "module_register_default_codechecker_tools",
+)
+use_repo(codechecker_extension, "default_codechecker_tools")
+register_toolchains("@rules_codechecker//src:codechecker_local_toolchain")
+```
+
+Or the following to your `WORKSPACE`:
+
+```python
+load(
+    "@rules_codechecker//src:tools.bzl",
+    "register_default_codechecker_tools",
+)
+
+register_default_codechecker_tools()
+register_toolchains(
+    "//src:codechecker_local_toolchain",
+)
+```
+
+### Providing your own tools
+
+If you don't want the default system tools -- for example to pin a specific
+CodeChecker version or point at a custom build —- you can define and register your own toolchain.
+
+First, define an implementation in a BUILD file with `codechecker_toolchain()`:
+
+```python
+load(
+    "@rules_codechecker//src:codechecker_toolchain.bzl",
+    "codechecker_toolchain",
+)
+
+codechecker_toolchain(
+    name = "codechecker_custom",
+    clang_tidy_binary  = "@my_tools//:clang-tidy",
+    clangsa_binary     = "@my_tools//:clang",
+    codechecker_binary = "@my_tools//:CodeChecker",
+)
+
+toolchain(
+    name = "codechecker_custom_toolchain",
+    toolchain = ":codechecker_custom",
+    toolchain_type = "@rules_codechecker//src:toolchain_type",
+    # Optionally constrain which execution platform this applies to:
+    # exec_compatible_with = ["@platforms//os:linux"],
+)
+```
+Finally register it in your `MODULE.bazel` (or `WORKSPACE`):
+```python
+register_toolchains("//path/to:codechecker_custom_toolchain")
+```
+
+[!NOTE]
+> Toolchains you register yourself take precedence over the default one if registered earlier.
--- a/README.md
+++ b/README.md
@@ -478,8 +478,6 @@ If you don't want the default system tools -- for example to pin a specific
 CodeChecker version or point at a custom build —- you can define and register your own toolchain.
 
 First, you will have to create or obtain labels for codechecker, clang and clang-tidy.
-In case you are unable to obtain a label for one, not defining the tool, will default to
-a target, pointing to the path given by `which <tool>`.
 
 Then define an implementation in a BUILD file with `codechecker_toolchain()`:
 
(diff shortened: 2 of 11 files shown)


src/codechecker_toolchain.bzl
```diff
@@ -0,0 +1,25 @@
+"""
+This file provides the toolchain rule for CodeChecker
+"""
+
+CodeCheckerInfo = provider(
+    doc = "This provider provides the executable path for CodeChecker",
+    fields = [
+        "executable",
+    ],
```
nettle: Tip: I think this should be a dict

src/codechecker_toolchain.bzl
```diff
@@ -0,0 +1,25 @@
+"""
+This file provides the toolchain rule for CodeChecker
+"""
+
+CodeCheckerInfo = provider(
+    doc = "This provider provides the executable path for CodeChecker",
```
nettle: I guess we should list all parameters including clang, clang-tidy and maybe more

src/codechecker_toolchain.bzl
```diff
@@ -0,0 +1,25 @@
...
+        codecheckerinfo = CodeCheckerInfo(
+            executable = ctx.attr.analyzer_binary,
+        ),
+    )
+    return [toolchain_info]
+
+codechecker_toolchain = rule(
+    implementation = _codechecker_toolchain_impl,
+    attrs = {
+        "analyzer_binary": attr.string(),
```
nettle: I guess "analyzer" is not a good name here

src/per_file.bzl
```diff
@@ -59,6 +59,8 @@ def _run_code_checker(
         content = "\n".join(ctx.attr.skip),
     )
 
+    info = ctx.toolchains["//src:toolchain_type"].codecheckerinfo
```
nettle, line 63: Yes, we discussed that to get rid of "src" you may try alias. (I havent tried myself though :) )

src/BUILD
```diff
@@ -63,3 +64,18 @@ label_flag(
...
     visibility = ["//visibility:public"],
 )
+
+# named this by convention
+# https://bazel.build/extending/toolchains#:~:text=%23%20By%20convention%2C%20toolchain%5Ftype%20targets%20are%20named%20%22toolchain%5Ftype%22%20and
+toolchain_type(name = "toolchain_type")
+
+codechecker_toolchain(
+    name = "codechecker_local",
+    analyzer_binary = "CodeChecker",  # Find it in PATH ?
```
nettle: We should find a way how to make it flexible. At least we should be able to pass CodeChecker binary by absolute path, try to find in the PATH (current implementation), as a Label, maybe via env var, maybe somehow else...

nettle (commented):
Good start! I think this is right way forward!

src/codechecker_toolchain.bzl
furtib: changed to codechecker_binary

src/codechecker_toolchain.bzl
furtib: I have added a way to define clang and clang-tidy.
For everything else, we can extend it later.

src/BUILD
furtib: I'm working on this.
I want to create targets with the repository rule and have those be the label() targets.

furtib: Fixed.

All binary targets are now labels.

README.md
```diff
@@ -427,3 +427,83 @@ After that you can find all artifacts in `bazel-bin` directory:
...
+load(
+    "@rules_codechecker//src:codechecker_toolchain.bzl",
+    "codechecker_toolchain",
+)
+
+codechecker_toolchain(
+    name = "codechecker_custom",
+    clang_tidy_binary  = "@my_tools//:clang-tidy",
+    clangsa_binary     = "@my_tools//:clang",
+    codechecker_binary = "@my_tools//:CodeChecker",
```
nettle: The description and examples are missing for `@my_tools//:CodeChecker` etc

src/codechecker_toolchain.bzl
```diff
@@ -0,0 +1,46 @@
+"""
+This file provides the toolchain rule for CodeChecker
+"""
+
+CodeCheckerInfo = provider(
+    doc = "This provider provides the executable path for CodeChecker and its related tools",
+    fields = {
+        "clang_tidy_bin": "clang-tidy executable",
```
nettle: Do we need suffix "_bin"? Does it have any meaning?

src/codechecker_script.py
```diff
@@ -29,7 +29,7 @@
 
 EXECUTION_MODE = "{Mode}"
 VERBOSITY = "{Verbosity}"
-CODECHECKER_PATH = "{codechecker_bin}"
+CODECHECKER_PATH = os.path.realpath("{codechecker_bin}")
```
nettle, line 32: So... we call `realpath()` only for CodeChecker?

src/codechecker_toolchain.bzl
```diff
@@ -0,0 +1,46 @@
...
+            clang_tidy_bin = ctx.executable.clang_tidy_binary,
+            clangsa_bin = ctx.executable.clangsa_binary,
+        ),
+    )
+    return [toolchain_info]
+
+codechecker_toolchain = rule(
+    implementation = _codechecker_toolchain_impl,
+    attrs = {
+        "clang_tidy_binary": attr.label(
```
nettle: The same question: "_binary" - do we need?

MODULE.bazel
```diff
@@ -38,3 +38,7 @@ codechecker_extension = use_extension(
     "module_register_default_codechecker_tools",
 )
 use_repo(codechecker_extension, "default_codechecker_tools")
+
+register_toolchains(
+    "//src:codechecker_local_toolchain",
```
nettle, line 43: I wonder what is the meaning of "locat"?
Is there a "global" alternative?

src/codechecker_toolchain.bzl
```diff
@@ -0,0 +1,46 @@
+"""
```
nettle, line 1: As we discussed, probably we do not need prefix "codechecker" in the file name.

nettle (commented):
I think this is the right direction!

MODULE.bazel
furtib: I named this local, since it uses the paths obtained by running `which *tool*` on the **local** machine.

The "global" (or more aptly named "remote") alternative, until we can write a version where Bazel manages all dependencies, should be written by end users, based on their specific remote executor setup.

src/codechecker_script.py
furtib: To create a target, I symlinked the binaries we used.
In the case of CodeChecker, it created problems with the virtual environment version, since the symlink was outside of the virtual environment. This is why I resolved the Path with `realpath`.

src/codechecker_toolchain.bzl
furtib: Removed

src/codechecker_toolchain.bzl
furtib: Also removed

src/codechecker_script.py
furtib: My bad, it seems we do need to resolve clang, and clang-tidy too.
Great catch!

README.md
furtib: Added one generic example.

src/per_file.bzl
```diff
@@ -59,11 +59,16 @@ def _run_code_checker(
...
```
Szelethus, line 63:
```suggestion
    # TODO: Consider using aliases so we don't have to type //src: everywhere.
    info = ctx.toolchains["//src:toolchain_type"].codecheckerinfo
```

src/codechecker_toolchain.bzl
Szelethus: I think while we have clang and clang-tidy standalone tools, this makes sense. This is not a hill I would die on, however, but this should be a separate PR if we want it.

Szelethus (approved):
LGTM. Lets not postpone pushing this any further -- I'm sure we'll need to adjust and fix things as we go on, but we need this in order to develop things incrementally.

PR #227, opened 2026-06-04
Remove aspect lint

furtib:
Why:
Aspect Lint does not provide any value over `buildifier_prebuilt`, but it causes compatibility problems.
The used aspect lint version is incompatible with `bazel 7.2.1`.

What:
- Removed aspect lint from dependencies
- Fixed buildifier version mismatch between MODULE and WORKSPACE systems.

Addresses:
Fixes: #229

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -21,7 +21,7 @@ bazel_dep(name = "rules_cc", version = "0.2.3")
 
 bazel_dep(
     name = "buildifier_prebuilt",
-    version = "6.4.0",
+    version = "7.3.1",
     dev_dependency = True,
 )
 single_version_override(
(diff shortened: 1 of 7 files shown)


Szelethus (approved):
I could reproduce the attached issue, and this PR indeed solves it. LGTM.

MODULE.bazel
```diff
@@ -21,7 +21,7 @@ bazel_dep(name = "rules_cc", version = "0.2.3")
 
 bazel_dep(
     name = "buildifier_prebuilt",
-    version = "6.4.0",
+    version = "7.3.1",
```
Szelethus, line 24: In the preceding PR you had issues with this, but `USE_BAZEL_VERSION=6.5.0 bazel test ...` runs fine for me.

furtib: The issue was not with running.
After a meeting, we decided to go ahead with this newer version.

PR #228, opened 2026-06-04
Update CI to follow how we run tests

furtib:
Why:
The way CI works and how tests are run locally (#218) shifted dramatically, so I updated the CI to correctly reflect how these tests were supposed to be run locally.

What:
- Merged lint/unit/FOSS CI into 1 (no reason to have them so fragmented)
- Used pytest to run Python tests
- Added `bazel test //...` to CI
- Replaced environment setup with micromamba
  Note:
    - In CI used a predefined micromamba action instead of our script to have caching available. This saves at least 2 minutes per job. 
    - Also, the `init.sh` script does not allow users to specify program versions with flags (e.g. `bazel=7.7.1`), which this other method does.
- Added a new `ci.yaml` micromamba environment. This was done so that the Bazel version is controllable with command-line arguments.
- Found hidden dependencies; these have been added to the `dev.yaml` file.
- Added a patch to the FOSS test yaml-cpp -- newer versions of gcc/clang failed to build the project

Addresses:
none

depends on: #233


.ci/micromamba/ci.yaml
```diff
@@ -0,0 +1,18 @@
+name: dev
```
Szelethus: Is there no way to merge this with `.ci/micromamba/dev.yaml`? In my view, its kind of the point of micromamba that we have a singular package declaration, which is used on all hosts to run the tests.

requirements.txt
```diff
@@ -1,3 +1,3 @@
 pytest
-pylint
+pylint >= 4.0.0
```
Szelethus: This is fine, but why the change?

test/foss/test_foss.py
```diff
@@ -82,6 +82,7 @@ def test_runner(self) -> None:
                 self.skipTest(
                     "This project is not compatible with this bazel version"
                 )
+            self.assertTrue(os.path.exists(os.path.join(test_dir,".git")))
```
Szelethus: Why is this important?

test/foss/yaml-cpp/yaml-cpp-070.patch
```diff
@@ -0,0 +1,9 @@
+--- a/BUILD.bazel
++++ b/BUILD.bazel
+@@ -10,5 +10,6 @@
+     visibility = ["//visibility:public"],
+     includes = ["include"],
+     hdrs = glob(["include/**/*.h"]),
++    copts = ["-include cstdint"],
```
Szelethus, line 8: Ahh, a famed patchfile!

Is it possible to include comments here?

test/foss/yaml-cpp/init.sh
```diff
@@ -38,6 +38,8 @@ git -C "$1" checkout yaml-cpp-0.7.0
 bazelversion="../../../.bazelversion"
 [ -f $bazelversion ] && cp $bazelversion "$1"
 
+patch $1/BUILD.bazel -p1 < yaml-cpp-070.patch
```
Szelethus, line 44: Comments?

.github/bazel_version.json
```diff
@@ -1,6 +1,6 @@
 [
     "6.5.0",
-    "7.7.0",
-    "8.5.0"
+    "7.7.1",
+    "8.5.1"
```
Szelethus: Why is this change necessary in this specific patch? Why not bump bazel 8 to up 8.6?

Szelethus (changes requested):
This patch, I think, is a little too large in scope.

requirements.txt
furtib: I had issues with older versions giving warnings while newer versions did not.
I thought the version was specified in README.md, turns out it wasn't.
To keep this PR lean, let's address this in another scope.

.ci/micromamba/ci.yaml
furtib: The original `dev.yaml` pinpoints a single Bazel version to use. (Therefore, it can be used easily)
Instead of creating duplicates for each Bazel version we want to test.

Added a comment to explain this.

test/foss/test_foss.py
furtib: It's not important (but good to have), left over from debugging FOSS tests.

Will remove to keep scope minimal.

test/foss/yaml-cpp/yaml-cpp-070.patch
furtib: It seems to be not entirely impossible but highly tool dependant...
Best if I keep comments in init.sh

.github/bazel_version.json
furtib: Micromamba didn't have 8.5.0 or 7.7.0. No reason for not updating straight to 8.6.

furtib: Yes, there is a reason:
The Bazel version 8.6.0 does not have the`bazel --version` command, which FOSS tests depend on.
To keep the scope of the patch minimal, this bump should be a separate PR.

Szelethus (approved):
LGTM

I don't really like that we created a new micromamba file. Please create an issue to merge them eventually by removing the bazel version specification in dev.yaml. Or come up with something even more clever. You were right all along...

PR #230, opened 2026-06-05
Fix timeout being 1000x longer than it should

furtib:
Why:
Currently, the timeout for checking if the CodeChecker server has started is not being divided by 1000, as it should be.

What:
- Fixed handling of milliseconds as seconds.

Addresses:
none


Szelethus (approved):
Nice catch!

PR #233, opened 2026-06-08
Detect analyzers not just in /bin for monolithic rule

furtib:
Why:
When running our tests, e.g. with micromamba, analyzers are not getting found due to PATH not being passed to the monolithic rule.
This went unnoticed so far, since `/bin` was being added to the path in the Python script.

What:
- Mimicking the way the codechecker binary is found, I find clang and clang-tidy too.
- Ran ormatter on `codechecker_script.py`

Addresses:
Fixes: #234

### DIFF
--- a/src/codechecker.bzl
+++ b/src/codechecker.bzl
@@ -135,7 +135,13 @@ def _codechecker_impl(ctx):
         # arguments = [ctx.outputs.codechecker_script.path],
         mnemonic = "CodeChecker",
         progress_message = "CodeChecker %s" % str(ctx.label),
-        # use_default_shell_env = True,
+        # While this is not an ideal solution
+        # until we have  better one (in the form of a toolchain)
+        # this is necessary to find binaries for analyzers
+        # installed in any place other than /bin. e.g.: micromamba
+        # /bin is an exception due to it being forcefully added
+        # in codechecker_script.py
+        use_default_shell_env = True,
     )
 
     # List all files required at build and run (test) time
--- a/src/codechecker.bzl
+++ b/src/codechecker.bzl
@@ -18,6 +18,8 @@ Rulesets for running codechecker in a single Bazel job.
 
 load(
     "@default_codechecker_tools//:defs.bzl",
+    "CLANG_BIN_PATH",
+    "CLANG_TIDY_BIN_PATH",
     "CODECHECKER_BIN_PATH",
 )
 load(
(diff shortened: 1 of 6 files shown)


src/codechecker.bzl
```diff
@@ -135,7 +135,13 @@ def _codechecker_impl(ctx):
         # arguments = [ctx.outputs.codechecker_script.path],
         mnemonic = "CodeChecker",
         progress_message = "CodeChecker %s" % str(ctx.label),
-        # use_default_shell_env = True,
+        # While this is not an ideal solution
+        # until we have  better one (in the form of a toolchain)
+        # this is necessary to find binaries for analyzers
+        # installed in any place other than /bin. e.g.: micromamba
+        # /bin is an exception due to it being forcefully added
+        # in codechecker_script.py
```
Szelethus: Here is how I would go about phrasing this, feel free to take as much or as little from it as you like.

"FIXME: Ideally, this action would run isolated of the invoking shell environment, using only the binaries bazel itself acquired. This would be best done by converting all tools (CodeChecker, clang, clang-tidy) into toolchains. Until then, we still need to allow the supply of these tools from PATH. For instance, micromamba (which we for testing) does not install to system folders like /bin, but uses PATH, making it impossible to configure binary version through it while use_default_shell_env is False."

src/codechecker.bzl
```diff
@@ -135,7 +135,18 @@ def _codechecker_impl(ctx):
...
         progress_message = "CodeChecker %s" % str(ctx.label),
-        # use_default_shell_env = True,
+        # FIXME: Ideally, this action would run
+        # isolated of the invoking shell environment,
+        # using only the binaries bazel itself acquired.
+        # This would be best done by converting all tools
+        # (CodeChecker, clang, clang-tidy) into toolchains.
+        # Until then, we still need to allow
+        # the supply of these tools from PATH.
+        # For instance, micromamba (which we for testing)
```
Szelethus:
```suggestion
        # For instance, micromamba (which we use for testing)
```

Szelethus (approved):
LGTM
@nettle I think this is a necessary evil we should consider living with. WDYT?

nettle (dismissed):
This is definitely not a fix.
And I'm not convinced that we should merge this workaround.
I suggest to look at line: `"{codechecker_bin}": CODECHECKER_BIN_PATH,`
I guess there is how the problem should be fixed or at least worked around.

Szelethus (commented):
Cheers @nettle for the input, this ended up looking much more ideal of a workaround.

I tested it out, analyzer version is correctly showing what micromamba set up!

```
bazel clean --expunge
source .ci/micromamba/init.sh
bazel test ...
cat $(find -L bazel-bin -type f -name "metadata.json") | jq | grep "\"version\""
```

PR #239, opened 2026-06-11
Fix per_file jobs not getting the codechecker tag

furtib:
Why:
per_file targets don't get the codechecker tag therefore, they are not excludable with the `--test_tag_filters=-codechecker` argument.

What:
- Added the codechecker tag to the per_file rule targets.

Addresses:
#238


Szelethus (approved):
LGTM

PR #240, opened 2026-06-11
Add warning about filtering

furtib:
Why:
Users are confused about why `--test_tag_filters=-codechecker` does not skip expensive analysis jobs.

What:
- Added a warning explaining how to correctly filter codechecker tests.

Addresses:
#238

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ bazel test ...
 ```
 
 > [!WARNING]
-> Filtering codecheckerr_tests with only `--test_tag_filters=-codechecker` is not enough.
+> Filtering codechecker_tests with only `--test_tag_filters=-codechecker` is not enough.
 > To skip the actual analysis `--build_tag_filters=-codechecker` must also be specified.
 
 You can find the analysis results in the `bazel-bin/` folder, on which you
--- a/README.md
+++ b/README.md
@@ -225,8 +225,8 @@ bazel test ...
 ```
 
 > [!WARNING]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -224,6 +224,10 @@ bazel test ://your_codechecker_rule_name
 bazel test ...
 ```
 
+> [!WARNING]
+> The option is still in prototype status and is subject to changes or removal without notice. See [#31](https://github.com/Ericsson/rules_codechecker/issues/31).
```
Szelethus: Are we sure that #31 is the issue to note here? Do we need to add any? Your PR is titled "Add warning about filtering", and this comment has little to do with filtering.

furtib: Whoops...
I forgot to save the file before commiting

README.md
```diff
@@ -224,6 +224,10 @@ bazel test ://your_codechecker_rule_name
 bazel test ...
 ```
 
+> [!WARNING]
+> Filtering codecheckerr_tests with only `--test_tag_filters=-codechecker` is not enough.
```
Szelethus:
```suggestion
> Filtering codechecker_tests with only `--test_tag_filters=-codechecker` is not enough.
```
Okay, this is great info,  but is this expected or desired behaviour?

furtib: Not desired, but users should know of it until we write a fix for it.

Szelethus (approved):
LGTM

PR #241, opened 2026-06-12
Disable workflow for changes made in documentation

furtib:
Why:
We don't need to run CI jobs for entirely documentation-related changes.

What:
- Disable CI jobs for pull request when only documentation was changed

Addresses:
none


Szelethus (approved):
Lets remember to re-enable them when we add any automation to documentation.

PR #242, opened 2026-06-16
Added --features=external_include_paths tests

furtib:
Why:
We want a reproducible error that we can fix.

What:
- Added tests for using `--features=external_include_paths` with both per_file and monolithic rule.

Addresses:
#236

### DIFF
--- a/test/unit/external_repository/test_external_repo.py
+++ b/test/unit/external_repository/test_external_repo.py
@@ -148,6 +148,8 @@ class TestImplDepExternalDep(TestBase):
 
     def test_codechecker_external_include_paths(self):
         """
+        Enabling external_include_paths will make bazel pass external  
+        include paths with -isystem.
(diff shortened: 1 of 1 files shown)


test/unit/external_repository/test_external_repo.py
```diff
@@ -145,6 +146,34 @@ def test_per_file_external_lib(self):
         )
         self.assertEqual(ret, 0, stderr)
 
+    def test_codechecker_external_include_paths(self):
+        """
+        Test: bazel test :codechecker_external_deps
```
Szelethus, line 153:
```suggestion
        Enabling external_include_paths will make bazel pass external
        include paths with -isystem.
        Test: bazel test :codechecker_external_deps
```

Szelethus (approved):
LGTM

PR #243, opened 2026-06-16
Fixed issues with  --features=external_include_paths

furtib:
Why:
We want to support running analysis with `--features=external_include_paths`.

What:
- Added flags from `external_includes` to compile commands.
- Added flags from `external_includes` of dependencies to compile commands.

Addresses:
Fixes: #236

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,8 +427,3 @@ After that you can find all artifacts in `bazel-bin` directory:
     
     # compile_commands.json for compile_commands_pass
     cat bazel-bin/test/compile_commands_pass/compile_commands.json
-
-
-Limitations:
-------------
-In Bazel 6, we do not support using the `--features=external_include_paths` flag.
--- a/README.md
+++ b/README.md
@@ -427,3 +427,8 @@ After that you can find all artifacts in `bazel-bin` directory:
     
     # compile_commands.json for compile_commands_pass
     cat bazel-bin/test/compile_commands_pass/compile_commands.json
+
+
+Limitations:
+------------
+In Bazel 6, we do not support using the `--features=external_include_paths` flag.
--- a/src/compile_commands.bzl
+++ b/src/compile_commands.bzl
@@ -112,11 +112,11 @@ def get_compile_flags(ctx, dep):
         if len(quote_include) == 0:
             quote_include = "."
         options.append(QUOTE_INCLUDE + quote_include)
-
-    for external_include in compilation_context.external_includes.to_list():
-        if len(external_include) == 0:
-            external_include = "."
-        options.append(EXTERNAL_INCLUDE + external_include)
+    if hasattr(compilation_context, "external_includes"):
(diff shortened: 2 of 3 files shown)


src/compile_commands.bzl
```diff
@@ -135,6 +141,16 @@ def get_compile_flags(ctx, dep):
                     system_include = "."
                 options.append(SYSTEM_INCLUDE + system_include)
 
+            for quote_include in compilation_context.quote_includes.to_list():
+                if len(quote_include) == 0:
+                    quote_include = "."
+                options.append(QUOTE_INCLUDE + quote_include)
```
Szelethus: This doesn't have a test.

Szelethus (changes requested):
Bazel 6.5.0 CI jobs break:
```
Error: 'CompilationContext' value has no field or method 'external_includes'
```
If this is a bazel 6 specific issue, then I suggest this might be a sign that it is time to officially stop supporting it.

README.md
```diff
@@ -427,3 +427,8 @@ After that you can find all artifacts in `bazel-bin` directory:
     
     # compile_commands.json for compile_commands_pass
     cat bazel-bin/test/compile_commands_pass/compile_commands.json
+
+
+Limitations:
+------------
+In Bazel 6, we do not support using the `--features=external_include_paths` flag.
```
furtib: I think it makes sense not to explicitly drop Bazel 6, and keep it in a limited, but working state.

Szelethus: I disagree. I don't want to half-support a bazel major version we already agreed on dropping.

furtib: Okay, let's drop Bazel 6 then.

src/compile_commands.bzl
furtib: I'm so far unable to create tests where I can influence quote_includes.

Szelethus: I suggest to move this to its own PR, because its not related, and it still requires tests. But we do want this fix!

test/unit/external_repository/test_external_repo.py
```diff
@@ -32,7 +32,7 @@ class TestImplDepExternalDep(TestBase):
     __test_path__ = os.path.dirname(os.path.abspath(__file__))
     BAZEL_BIN_DIR = os.path.join("bazel-bin")
     BAZEL_TESTLOGS_DIR = os.path.join("bazel-testlogs")
-    BAZEL_VERSION = None
+    BAZEL_VERSION: str = ""
```
furtib, line 35: This fixes type warnings

src/compile_commands.bzl
furtib: I have moved this part into its own PR: #255.

Szelethus (commented):
I meant to submit this earlier, but you have already fixed it.

Szelethus (approved):
LGTM

PR #244, opened 2026-06-16
Remove bazel 6 from CI and testing

furtib:
Why:
We have previously agreed that as long as we keep the WORKSPACE files working, we can drop support for bazel 6.
Since we encountered issues where Bazel 6 compatibility can only support a subset of features, we decided not to maintain Bazel 6 in a half-working state but rather drop it entirely.

What:
- Removed bazel 6 from CI

Addresses:
none


.github/bazel_version.json
```diff
@@ -1,5 +1,4 @@
 [
-    "6.5.0",
```
nettle, line 2: I'm sure we do not need this file at all :)

MODULE.bazel
```diff
@@ -14,7 +14,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-module(name = "rules_codechecker")
+module(
+    name = "rules_codechecker",
+)
```
nettle: Any reason for this change?

Szelethus (approved):
LGTM

PR #245, opened 2026-06-17
Limit/declare thread usage to bazel

furtib:
Why:
Bazel expects its actions to use 1 thread by default.
This can result in it scheduling multiple long analysis jobs on the same machine, each trying to hog resources from the others.

What:
- Limit thread usage of per_file to 1
- Try declaring thread usage for the monolithic rule (only affects local execution)

Addresses:
#246

### DIFF
--- a/src/codechecker.bzl
+++ b/src/codechecker.bzl
@@ -43,7 +43,7 @@ load(
     "per_file_test",
 )
 
-def _codechecker_resource_set(os_name, input_count):
+def _codechecker_resource_set(_os_name, input_count):
     """
     Requests thread count based on number of input files.
 
@@ -52,7 +52,7 @@ def _codechecker_resource_set(os_name, input_count):
     This may not work with remote machines.
     """
     return {
-        "cpu": input_count, # analysis is run for most input files
+        "cpu": input_count,  # analysis is run for most input files
     }
 
 def get_platform_alias(platform):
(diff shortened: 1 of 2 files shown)


Szelethus (commented):
I'd appreciate some more explanation, preferably in an issue on what problem we are solving here. I'll do my best to do my own research on this!

src/per_file_script.py
```diff
@@ -93,6 +93,7 @@ def _run_codechecker() -> None:
         + ["--skip", SKIP_FILE]
         + ["--config", CONFIG_FILE]
         + [COMPILE_COMMANDS_ABSOLUTE]
+        + ["-j 1"] # Analysis of a single file should use a single thread!
```
nettle, line 96: I guess this has no much sense for per-file :)
And I would not recommend doing changes "just for the case" too.

src/codechecker.bzl
```diff
@@ -43,6 +43,18 @@ load(
...
+    """
+    Requests thread count based on number of input files.
+
+    If requested thread count is higher than available
+    bazel schedules the job to run alone.
+    This may not work with remote machines.
+    """
+    return {
+        "cpu": input_count,  # analysis is run for most input files
+    }
```
nettle, line 56: Interesting! And this may actually help to improve performance!
But we should be sure... Need measurement results

PR #248, opened 2026-06-18
Convert unit tests into the new unit_test macro

furtib:
Why:
We want our tests to run with `bazel test //...`.

What:
- Rewritten every test where possible as the new unit_test macro

Addresses:
#210


src/per_file.bzl
```diff
@@ -165,9 +165,9 @@ def _per_file_impl(ctx):
         fail("Seems compile_commands.json file is incorrect!")
     sources_and_headers = _collect_all_sources_and_headers(ctx)
     options = ctx.attr.default_options + ctx.attr.options
-    all_files = [compile_commands]
     config_file, env_vars = get_config_file(ctx)
     _create_wrapper_script(ctx, options, compile_commands, config_file)
+    all_files = [compile_commands, config_file]
```
Szelethus: Considering the nature of this PR, it would be amazing to land this separately and make this NFC.

Szelethus (commented):
I haven't read the PR in its entirety, just a preliminary comment.

src/per_file.bzl
furtib: This was done in a separate PR already, but has not been rebased yet.
(After rebase, this line was duplicated, despite no merge conflicts)

test/unit/generated_files/BUILD
```diff
@@ -114,3 +118,52 @@ compile_commands(
...
+    name = "codechecker_genrule_source_test",
+    contains = ["defect(s) in genrule_source.cc"],
+    data = [":codechecker_genrule_source"],
+    files = "test/unit/generated_files/codechecker_genrule_source/codechecker.log",
+)
+
+unit_test(
+    name = "per_file_genrule_header_test",
+    contains = [
+        "genrule_header.h",
```
Szelethus, line 139: The original test checks for `"defect\(s\) in genrule_header.h"`

test/unit/generated_files/BUILD
```diff
@@ -114,3 +118,52 @@ compile_commands(
...
+    ],
+    data = [":per_file_genrule_header"],
+    excludes = "genrule_source.cc",
+    files = "test/unit/generated_files/per_file_genrule_header/data/test-unit-generated_files-genrule_header_consumer.cc_clangsa.plist",
+)
+
+unit_test(
+    name = "per_file_genrule_source_test",
+    contains = [
+        "genrule_source.cc",
```
Szelethus, line 150: same here

test/unit/virtual_include/test_virtual_include.py
```diff
@@ -1,139 +0,0 @@
...
-        )
-        # Test whether the _virtual_include directory was actually created.
-        self.assertTrue(
-            os.path.isdir(f"{self.BAZEL_BIN_DIR}/_virtual_includes")
-        )
-        # FIXME: In the postprocessed plists, all _virtual_include paths
-        # should've been removed. Possible fix is in the github PR #14.
-        self.assertNotEqual(
-            self.contains_in_files(r"/_virtual_includes/", plist_files), []
-        )
```
Szelethus, line 93: Is this present in the new tests?

Szelethus (changes requested):

test/unit/generated_files/BUILD
furtib: The original test checks in the `test.log` file, which we cannot test here.
We can check for that in the monolithic rules case because the rule outputs the equivalent of this log (result of `CodeChecker parse`) as a build artifact.
This is the closest alternative I could find that works for plist files.

test/unit/virtual_include/test_virtual_include.py
furtib: Yes, in the `codechecker_plist_path_resolved_test` test on line 106.

Szelethus (approved):
LGTM

PR #249, opened 2026-06-18
Fix config file not being part of DefaulInfo files

furtib:
Why:
We want config files as output files for our rule.
We use the config file in testing: `test/unt/config`.
The current setup works, but is not explicitly supported by Bazel.
Using remote execution or rewriting the test to run inside Bazel results in the file not being found for the tests.

What:
- Added config files to `DefaultInfo` files output.

Addresses:
none


Szelethus (approved):
LGTM

PR #250, opened 2026-06-18
Update yaml-cpp and remove patchfile

furtib:
Why:
Having to patch our test is not desirable.
The effect the patch achieves was done in a newer release of yaml-cpp, let's use that instead.

What:
- Bumped yaml-cpp version to 0.8.0
- Removed patchfile for yaml-cpp-0.7.0
- Removed patch dependency from micromamba

Addresses:
none


Szelethus (approved):
LGTM

PR #252, opened 2026-06-29
Remove python toolchain

furtib:
Why:
The situation around the Python toolchain is very complicated.
We use a self-created Python toolchain as well as the official rules_python toolchain.
We do not use either toolchain at all, so to keep it simple (KISS), let's remove them completely.
If we really do need it (e.g., to obtain packages through pip with rules_python), we should add it back in the PR where it becomes necessary.

What:
Removed every Python toolchain from the project.

Addresses:
Fixes: #251


src/BUILD
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Tool filter compile_commands.json file
-# In bazel 6 we use our own python toolchain,
+# In bazel 6 we use the built in python toolchain,
```
Szelethus: We no longer use bazel 6, this is worth updating.

Szelethus (commented):
Surprising. @nettle, any thoughts?

Szelethus (approved):
I've been thinking on this for a little while, and I don't see a downside to this. If we find out later that we did need this, let's land this back *with tests*.

LGTM

nettle (approved):
I've briefly tested with my project - seems fine

src/BUILD
```diff
@@ -14,7 +14,7 @@
 load(":codechecker_toolchain.bzl", "codechecker_toolchain")
 
 # Tool filter compile_commands.json file
-# In bazel 6 we use our own python toolchain,
+# In bazel 6 we use the built in python toolchain,
```
Szelethus:
```suggestion
# We use the built in python toolchain,
```

PR #253, opened 2026-06-29
Use py_binary with constants file, instead of arguments

furtib:
Why:
I want to create a more organized, easily understandable alternative to #199.
The goal is to choose one of these two.

Since there is a limit on the length of the arguments we can pass to our scripts, and passing many arguments makes things complicated, I decided to replace `expand_template` with a `bazel_constants.json` file.
This way, we can pass any number of arguments in an organized manner to the Python script.

What:
- Replaced the functionality of expand_template with a bazel_constant.json file for the per_file rule. 

Addresses:
none

### DIFF
--- a/src/BUILD
+++ b/src/BUILD
@@ -23,11 +23,21 @@ py_binary(
     visibility = ["//visibility:public"],
 )
 
+# In bazel 6 we use our own python toolchain,
+# therefore we cannot load from rules_python.
+# buildifier: disable=native-py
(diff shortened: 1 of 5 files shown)


Szelethus (commented):
I'm not convinced by the superiority of this solution against #199. For one, constructing a json file for testing purposes sounds more troublesome.

What is your view on this? Which one would you prefer to see landed?

PR #254, opened 2026-06-30
Add test for metadata

furtib:
Why:
We want to test metadata contents before merging #79.

What:
- Added tests for inspecting the contents of metadata files.

Addresses:
#45


test/unit/metadata/BUILD
```diff
@@ -0,0 +1,68 @@
...
+    srcs = [
+        "main.cc",
+        "other.cc",
+    ],
+    tags = ["manual"],
+)
+
+codechecker_test(
+    name = "codechecker_multiple_source",
+    tags = ["manual"],
```
Szelethus, line 40: Lets mark explicitly that this is `per_file = False`. Also, the name should be changed to reflect this very important difference.

test/unit/metadata/BUILD
```diff
@@ -0,0 +1,68 @@
...
+    name = "codechecker_metadata_test",
+    contains = [r"\"action_num\": 2,"],
+    data = [":codechecker_multiple_source"],
+    files = "test/unit/metadata/codechecker_multiple_source/codechecker-files/data/metadata.json",
+)
+
+unit_test(
+    name = "per_file_metadata_test",
+    data = [":per_file_multiple_source"],
+    excludes = [r"\"action_num\": 2,"],
```
Szelethus: I suppose this is the important bit, no? That this *should* be `contains` instead.

Szelethus (changes requested):

test/unit/metadata/BUILD
furtib: Great catch, fixed it!

test/unit/metadata/BUILD
```diff
@@ -0,0 +1,69 @@
...
+    files = "test/unit/metadata/monolithic_multiple_source/codechecker-files/data/metadata.json",
+)
+
+unit_test(
+    name = "per_file_metadata_test",
+    contains = [r"\"action_num\": 2,"],
+    data = [":per_file_multiple_source"],
+    files = "test/unit/metadata/per_file_multiple_source/data/metadata.json",
+    # FIXME: remove when metadata is generated
+    tags = ["manual"],
```
Szelethus: As is, I don't think this is a great precursor patch. We could check whether this (these?) files exists at all, AND THEN check how merging works.

furtib: There is currently no way to check if a file is missing.
If no input file is found the test just fails.

furtib: I tried to make a way to assert files existing #261.

Using that, we could assert the existence of individual metadata files, but not the merged one.

Szelethus (commented):
LGTM if you rebase on #261 and adjust the test.

Szelethus (approved):
LGTM

PR #255, opened 2026-06-30
Add transitive quote includes to compile commands

furtib:
Why:
We want to see quotes include paths from transitive dependencies.

What:
- Added paths from quote includes of dependencies.

Addresses:
none


Szelethus (changes requested):
As we discussed, this is quite trivial, but we kinda need tests anyways.

furtib:
Superseded by: #272.

PR #256, opened 2026-07-04
Add default values for clang, clang-tidy etc. in toolchain

furtib:
Why:
We do not/cannot expose the clang, clang-tidy, codechecker, etc. targets to users, so when they want to override only one of these, they need to provide targets for each.
To make it easier, we can expose our binary targets by making them the default value for the toolchain.

What:
- Added a default value to each binary provided by our toolchain. 

Addresses:
none


Szelethus (approved):
Looks reasonable.

PR #257, opened 2026-07-04
Fix visibility of toolchain target

furtib:
Why:
No one can create a new toolchain target due to its restrictive visibility.

What:
- Made the toolchain_type target visible from outside the rules_codechecker codebase.

Addresses:
none


Szelethus (approved):
LGTM

PR #258, opened 2026-07-05
Add CI job to force WORKSPACE usage

furtib:
Why:
We currently have no CI workflow using our workspace file.
Since Bazel 7 (the last version that) supports WORKSPACE files, we should test it working with just the workspace file too, not just with the MODULE.bazel file. 

What:
- Add new workflow for bazel 7.7.1 forcing workspace usage.

Addresses:
none

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -27,7 +27,6 @@ on:
       - "README*"
       - "LICENSE"
 
-
 permissions: read-all
 
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -36,14 +35,20 @@ concurrency:
...
-          - "8.5.1"
+        include:
+          - bazel_version: "7.7.1"
+            bazel_mode: "workspace"
+
+          - bazel_version: "7.7.1"
+            bazel_mode: "module"
+
+          - bazel_version: "8.5.1"
+            bazel_mode: "module"
```
Szelethus, line 51: This looks trivial from in the context of the PR, but lets rationalize this with a few comments.

PR #260, opened 2026-07-05
feat: distributed rule metadata files

### LINKED ISSUE #45: Generate metadata.json in file-by-file rule

furtib:
Why:
We don't need multiple metadata files, just one.

What:
- Merge metadata files.

Addresses:
Fixes: #45 

Depends on:
#79

### DIFF
--- a/src/metadata_merge.py
+++ b/src/metadata_merge.py
@@ -1,0 +1,124 @@
+# Copyright 2023 Ericsson AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Merges the metadata information of multiple CodeChecker analysis.
+"""
+
+import json
+import os
+import sys
+from typing import List, Dict, Any
+
+
+def merge_two_json(json1, json2):
+    """
+    Merges the data of two json files.
+
+    If both json files are empty, returns an empty json.
+    """
+    if json1 == {}:
+        return json2
+    if json2 == {}:
+        return json1
+    # Happens when analysis of all files was skipped
+    if json1 == {} and json2 == {}:
+        return {}
+    # Fail if the plist file version is different
+    assert json1["version"] == json2["version"]
+    json1_root = json1["tools"][0]
+    json2_root = json2["tools"][0]
+    # Command, working directory nad output directory may be different
+    # from metadata to metadata, due to remote workers.
+    # Currently we choose the first of these values.
+    # We expect the following fields to be the same in all metadata files.
+    assert json1_root["name"] == json2_root["name"]
+    # same CodeChecker version
+    assert json1_root["version"] == json2_root["version"]
+    # We assume that the list of enabled checkers haven't changed between runs.
+    # We append info from json2 to json1 from here on out
+    json1_root["action_num"] += json2_root["action_num"]
+    json1_root["result_source_files"].update(json2_root["result_source_files"])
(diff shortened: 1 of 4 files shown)


src/metadata_merge.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    if json1 == {}:
+        return json2
+    if json2 == {}:
+        return json1
+    # Happens when analysis of all files was skipped
+    if json1 == {} and json2 == {}:
+        return {}
+    # Fail if the plist file version is different
+    assert json1["version"] == json2["version"]
```
Szelethus: I'll one up you: Fail if the version isn't 2.

Szelethus (changes requested):
Please note that the structure of metadata.json, specifically version 2, is specified here:
https://github.com/Ericsson/codechecker/blob/master/docs/report_directory.md#metadata-structure

It might be wise to drop this somewhere in the new metadata_merge.py file.

src/metadata_merge.py
```diff
@@ -0,0 +1,124 @@
...
+    json1_root = json1["tools"][0]
+    json2_root = json2["tools"][0]
```
Szelethus: `*_root` sounds like a poor name. How about `*_first_tool` or something like that?

src/metadata_merge.py
```diff
@@ -0,0 +1,124 @@
...
+    assert json1_root["version"] == json2_root["version"]
+    # We assume that the list of enabled checkers haven't changed between runs.
+    # We append info from json2 to json1 from here on out
+    json1_root["action_num"] += json2_root["action_num"]
+    json1_root["result_source_files"].update(json2_root["result_source_files"])
+    json1_root["skipped"] = json1_root["skipped"] + json2_root["skipped"]
+    # Merge time; we assume here both json files describe jobs in
+    # the same analysis invocation, implying that the analysis start
+    # time is the lowest timestamp, and the end is the highest.
+    # Note: caching will break this assumption
```
Szelethus, line 137: That sounds like a bigger deal than this Note implies.

src/metadata_merge.py
```diff
@@ -0,0 +1,124 @@
...
+    json1_root = json1["tools"][0]
+    json2_root = json2["tools"][0]
+    # Command, working directory nad output directory may be different
+    # from metadata to metadata, due to remote workers.
+    # Currently we choose the first of these values.
+    # We expect the following fields to be the same in all metadata files.
+    assert json1_root["name"] == json2_root["name"]
+    # same CodeChecker version
+    assert json1_root["version"] == json2_root["version"]
+    # We assume that the list of enabled checkers haven't changed between runs.
```
Szelethus: What if we have several, differently configured `codechecker_test` targets, on which a main test target depends? Would each of those targets do their metadata merge individually?

Can we (should we) assert on this?

Szelethus (changes requested):
It'd be nice to follow the official documents order of items.

Also, we could have a very nice unit test on this. And currently we have none :)

src/metadata_merge.py
furtib: The merging happens on a per-codechecker_test target basis, to mimic/follow the behavior of the non per-file codechecker_test rule.
I have written functions to merge analyzer sections of metadata files too.
Is this assert still relevant in this case?

src/metadata_merge.py
furtib: I extended the Note, not sure what else could I do here.

src/metadata_merge.py
furtib: Renamed.

PR #261, opened 2026-07-05
Extend unit test with the ability to assert certain files not existing

furtib:
Why:
We want to be able to write tests that assert that certain files do not exist (and require that all files listed under files exist).

What:
- Added parameter to assert that a file does not exist (if these files are also in files, remove them)
- Made it a requirement for all `file = ["..."]` patterns to match a file.

Addresses:
none

### DIFF
--- a/test/unit/grep_check.py
+++ b/test/unit/grep_check.py
@@ -148,7 +148,7 @@ def check_file(content: str, args) -> tuple[bool, set[str], set[str]]:
     return all_passed, found_patterns, missing_patterns
 
 
-def assert_files_missing(files):
+def assert_expectedly_missing_files(files):
     """
     Asserts that none of the files exist in `files`.
     """
--- a/test/unit/grep_check.py
+++ b/test/unit/grep_check.py
@@ -45,7 +45,7 @@ def parse_args() -> argparse.Namespace:
         help="Path or glob pattern to the file(s) to search within.",
     )
     parser.add_argument(
-        "--no_files",
+        "--missing_files",
         nargs="+",
         required=False,
         help="Path or glob pattern to the file(s) that shouldn't exists.",
@@ -187,9 +187,9 @@ def main() -> None:
     args = parse_args()
     check_args(args)
 
-    assert_files_missing(args.no_files)
+    assert_expectedly_missing_files(args.missing_files)
 
-    files_to_check = [x for x in args.files if x not in args.no_files]
+    files_to_check = [x for x in args.files if x not in args.missing_files]
     if not files_to_check:
         print(
             "There is no files to check, "
--- a/test/unit/grep_check.py
+++ b/test/unit/grep_check.py
@@ -166,6 +166,22 @@ def fail(msg: str, exit_code: int = 1):
     print(msg)
     sys.exit(exit_code)
 
+
+def collect_file_paths(glob_list: list[str]) -> list[str]:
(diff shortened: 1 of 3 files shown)


test/unit/grep_check.py
```diff
@@ -143,25 +148,60 @@ def check_file(content: str, args) -> tuple[bool, set[str], set[str]]:
     return all_passed, found_patterns, missing_patterns
 
 
-def main() -> None:
-    """Entry point for the pattern-matching test."""
-    args = parse_args()
-    check_args(args)
+def assert_files_missing(files):
+    """
+    Asserts that the list of file paths given as parameter, doesn't exists.
```
Szelethus:
```suggestion
    Asserts that none of the files exist in `files`.
```

test/unit/grep_check.py
```diff
@@ -143,25 +148,60 @@ def check_file(content: str, args) -> tuple[bool, set[str], set[str]]:
...
+        fail("  [ERR] No file collected to be checked.")
+    return file_paths
+
+
+def main() -> None:
+    """Entry point for the pattern-matching test."""
+    args = parse_args()
+    check_args(args)
+
+    assert_files_missing(args.no_files)
```
Szelethus: I've been looking at this PR for a bit, and this is the point that convinced me that "absent" isn't the correct phrase. You write "missing" here, but those files are "absent", but that doesn't really telegraph whats desiren, which is "expectedly missing" or "should be missing".

test/unit/grep_check.py
```diff
@@ -143,25 +148,60 @@ def check_file(content: str, args) -> tuple[bool, set[str], set[str]]:
...
-    args = parse_args()
-    check_args(args)
+def assert_files_missing(files):
+    """
+    Asserts that the list of file paths given as parameter, doesn't exists.
+    """
+    for file_pattern in files:
+        matched_files = glob.glob(file_pattern, recursive=True)
+        if matched_files:
+            print(f"\n{file_pattern} exists, however it shouldn't.")
```
Szelethus: "unexpectedly exists"

Szelethus (commented):
I know I'm nitpicking, but the phrasing is off. The overall PR direction is good.

test/unit/grep_check.py
furtib: renamed it to `expected_missing_file`.

test/unit/grep_check.py
```diff
@@ -44,6 +44,12 @@ def parse_args() -> argparse.Namespace:
         required=True,
         help="Path or glob pattern to the file(s) to search within.",
     )
+    parser.add_argument(
+        "--no_files",
```
Szelethus: Can we adjust this name as well?

test/unit/grep_check.py
```diff
@@ -143,25 +148,60 @@ def check_file(content: str, args) -> tuple[bool, set[str], set[str]]:
     return all_passed, found_patterns, missing_patterns
 
 
-def main() -> None:
-    """Entry point for the pattern-matching test."""
-    args = parse_args()
-    check_args(args)
+def assert_files_missing(files):
```
Szelethus:
```suggestion
def assert_expectedly_missing_files(files):
```

Szelethus (approved):
LGTM after some naming adjustments.

PR #262, opened 2026-07-08
Added ZSH support for micromamba

Makkareal:
Why:
When using Zsh source ` source .ci/micromamba/init.sh` did not work, because it did not find the location of the script.

What:
Added a branch if the shell is Zsh

Addresses:
None

### DIFF
--- a/.ci/micromamba/init.sh
+++ b/.ci/micromamba/init.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 # Make sure we source the script
 if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
     echo "ERROR: This script must be 'sourced':"
--- a/.ci/micromamba/init.sh
+++ b/.ci/micromamba/init.sh
@@ -38,7 +38,13 @@ ENV_NAME="${1:-dev}"
 log "Environment: $ENV_NAME"
 
 # Get the location of this script
-THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+if [ -n "$ZSH_VERSION" ]; then
+    RAW_SOURCE="${(%):-%x}"
+else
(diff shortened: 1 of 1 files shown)


.ci/micromamba/init.sh
furtib, line 3: This line does not work in ZSH; see the following example script:
```zsh
furtib@emypc ~ % cat test.sh 
if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
    echo "ERROR: This script must be 'sourced':"
    echo "       source ${0}"
    exit 1
fi


furtib@mypc ~ % ./test.sh
./test.sh: 1: Bad substitution
```

furtib (changes requested):
Hello, welcome to the project!

There are multiple bash script files using this or similar methods for obtaining the location of the script (e.g., `.ci/micromamba/run_test.sh`, `.ci/micromamba/uninstall.sh`); those should be updated too.
If possible, I suggest avoiding the ZSH-specific execution path and using something that works universally. (~The solution I used in run_test.sh seems to work in ZSH too, please check that!~ -- does not work with sourcing)

I'm slightly concerned about the impact of ZSH support inside basel.

Since we are not explicitly supporting ZSH, I relabeled this as an enhancement, not a bug.

.ci/micromamba/init.sh
Makkareal: Added shebang, so the script always runs in bash when not sourced.

Unfortunately, I could not find a universal way of getting the correct path. If the support of ZSH is a concern, then perhaps it would be better to simply check if the script is sourced from bash and exit with an instruction if it is not.

I will update the other scripts with whichever solution is preferred.

Szelethus: That sounds like a great alternative solution. Lets pursue that.

Szelethus (changes requested):
In hindsight, I also think it'd be wiser to only support bash.

PR #264, opened 2026-07-31
Document CodeChecker skip limitations

### LINKED ISSUE #193: Document skip limitations
### Describe the task

The skip argument cannot contain absolute paths since CodeChecker is run in the sandbox environment.
The per_file rule for now doesn't support the skip argument at all. (The per_file rule won't be able to use CodeCheckers skipfile implementation.)

### Scope

- Update documentation with skip argument.

### Definition of Done

- Documentation updated

lntutor:
Why:
The `skip` argument was not documented, so users could reasonably try absolute host paths that cannot match files inside Bazel's sandbox.

What:
- document the CodeChecker skip-file syntax
- explain why workspace-relative patterns are required
- add an example matching the repository's skip tests
- note that current standard and per-file analysis both support `skip`

Validation:
- `git diff --check`
- compared the example with `test/unit/skip/BUILD`
- verified both rule implementations pass `skip` to CodeChecker

Addresses:
Fixes #193

🤖 Generated with AI

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,29 @@ codechecker_test(
 )
 ```
 
+#### Skipping files
+
+The `skip` argument accepts a list of patterns using the
+[CodeChecker skip-file syntax](https://codechecker.readthedocs.io/en/latest/analyzer/user_guide/#skip-file).
+Because CodeChecker runs inside the Bazel sandbox, skip patterns cannot use
+absolute paths from the host system. Use workspace-relative patterns instead,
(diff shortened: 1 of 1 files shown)


furtib (changes requested):
This addition is factually correct; however, you mention the per-file analysis before it is introduced.

Please move the section right before the section titled `Multi-platform CodeChecker analysis: codechecker_suite()`.

furtib (approved):
LGTM!

PR #266, opened 2026-08-03
Make per_file test script use the codechecker toolchain

furtib:
Why:
The test script (that runs `CodeChecker parse`) for the per_file rule does not use the CodeChecker toolchain.
Also reproducible by using Bazel 9 (If load statements are correctly placed).

What:
- Rewrote the per_file test script to use the CodeChecker, specified by the toolchain.

Addresses:
Fixes: #268


Szelethus (approved):
LGTM. I understand its pretty hard to write a test for this; is it quite impossible?

PR #267, opened 2026-08-05
Make bazel version query more robust

furtib:
Why:
Over and over, we encounter a problem with the current version of querying the bazel version, resulting in failures.
We have already updated the method to use `bazel version --gnu_format` in the external unit test.
The relevant code snippet:
https://github.com/Ericsson/rules_codechecker/blob/f39b2687a8b6567e5b5ca30629308565d9b60091/test/unit/external_repository/test_external_repo.py#L60-L66

What:
- Made bazel version detection more robust in FOSS tests.

Addresses:
none


Szelethus (approved):
ezpz LGTM

PR #269, opened 2026-08-06
Introduce public API of rules_codechecker

nettle:
Why:
Users should not load rules from the implementation files under src/,
which exposes internals and deviates from the Bazel convention
of a single defs.bzl entry point per rules repository.

What:
- Add defs.bzl exporting the user facing rules and macros
- Switch README examples to @rules_codechecker//:defs.bzl
- Switch unit and foss test BUILD files to the new load path
- Keep clang_ctu_test and the internal codechecker rule in src/

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ To use `codechecker_test()` include it to your BUILD file:
 
 ```python
 load(
-    "@rules_codechecker//src:codechecker.bzl",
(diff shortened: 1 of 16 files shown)


furtib (approved):
LGTM!
I don't know why CI hasn't run for this PR either; I will investigate.

PR #270, opened 2026-08-06
Remove WORKSPACE files

nettle:
Why:
Bazel 8 disables the WORKSPACE by default and Bazel 9 removed it.
Our WORKSPACE file is not used FOSS tests were skipped on Bazel 8,
which left yaml-cpp untested there.

What:
- Delete WORKSPACE and test/foss/templates/WORKSPACE.template
- Point the buildifier at MODULE.bazel
- Remove zlib-module FOSS test
- Convert yaml-cpp FOSS test to bzlmod, pin tag yaml-cpp-0.9.0
- Drop the bazel 8 skip and the WORKSPACE handling from the FOSS tests
- Document only the MODULE.bazel


furtib (approved):
If we do not need to support WORKSPACE files, I'm all for this change!

Keep in mind that CI has not run for whatever reason!

I was hesitant to take this step due to Bazel 7 still supporting WORKSPACE files.
If we merge this don't forget to close: #258.

PR #271, opened 2026-08-07
Refactor micromamba init

nettle:
Why:
We need to run different tool version combinations in CI and locally.
The versions were hardcoded in two nearly identical environment files,
and init.sh could only create and activate that one fixed environment.

What:
- Generate env file from env.yaml template
- Remove dev.yaml, superseded by the template
- Add --bazel, --codechecker and --force options to init.sh
- Create the environment when executed, activate when sourced
- Reuse an existing environment
- Print versions of tools
- Split init.sh into functions, keep its variables out of the shell
- Refactor run_tests.sh


furtib (commented):
Program versions in micromamba can be overwritten by command-line arguments if not specified by the YAML file.
Like so: `micromamba create --file ci.yaml --name myenv bazel=8.5.1`.
Wouldn't it be better to use micromamba this way instead of creating a separate configuration file for each environment?

furtib (approved):
I mean LGTM 😅

PR #273, opened 2026-08-10
Use bazel's built in symlink instead of cp for config files

furtib:
Why:
The `codechecker_config` rule uses `cp` to duplicate the user-provided config file. This is an unnecessary dependency on `cp`, and may not work correctly with remote executors.
Using `ctx.actions.symlink` is the idiomatic Bazel approach, so it should work with remote execution and remote caching.

What:
- Replace the `ctx.actions.run(executable = "cp", ...)` call in `src/codechecker_config.bzl` with `ctx.actions.symlink(output, target_file)`.

Addresses:
none

### DIFF
--- a/src/codechecker_config.bzl
+++ b/src/codechecker_config.bzl
@@ -63,19 +63,13 @@ def get_config_file(ctx):
         else:
             config_info = ctx.attr.config[CodeCheckerConfigInfo]
         if config_info.config_file:
-            # Create a copy of CodeChecker configuration file
+            # Symlink CodeChecker configuration file
             # provided via codechecker_config(config_file)
             config_file = config_info.config_file.files.to_list()[0]
-            ctx.actions.run(
-                inputs = [config_file],
-                outputs = [ctx_config_file],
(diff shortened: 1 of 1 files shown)


src/codechecker_config.bzl
```diff
@@ -63,19 +63,13 @@ def get_config_file(ctx):
...
-                progress_message = "Copying CodeChecker config file",
-                executable = "cp",
-                arguments = [
-                    config_file.path,
-                    ctx_config_file.path,
-                ],
+            ctx.actions.symlink(
+                output = ctx_config_file,
+                target_file = config_file,
+                progress_message = "Symlinking CodeChecker config file",
```
nettle, line 72: Great idea!

nettle (commented):
I guess this might be the best alternatives (among /usr/sbin/cp, no-remote, create file etc)!
The toughest question is the test...

Szelethus (approved):
We already recieved word from a user who has a working BuildBarn instance, and this PR seems to fix the missing cp binary issue.

PR #274, opened 2026-08-10
Add dependencies of tools from toolchain to runfiles

### LINKED ISSUE #292: Hermetic codechecker cannot be used with toolchain
**Describe the bug**
CodeChecker hermetically provided to Bazel cannot be used because its runfiles (files necessary for it to run) are not collected by the toolchain.

**Versions**
* CodeChecker 6.28.2
* bazel 8.5.1
* bed72c8b2efd0cc5ac3ab37840a15656ffa200df

**To Reproduce**
Run the following script below. cd into the created `repro` directory and run `bazel test //...`. You will see that codechecker is not found.

To test the fixed behaviour (#274), modify the `MODULE.bazel` file. (uncomment lines 16-17 and comment lines 13-14)

Short description of the script:
- creates a standalone compilation of codechecker
- uses the obtained tar.gz to obtain codechecker inside Bazel

```bash
#!/bin/bash
set -e

WORKDIR="${1:-repro}"
WORKDIR_CC="${2:-repro_cc_package}"

WORKDIR="$(realpath -m "$WORKDIR")"
WORKDIR_CC="$(realpath -m "$WORKDIR_CC")"

if [ -d "$WORKDIR" ]; then
    echo "Directory '$WORKDIR' already exists. Remove it or choose another name."
    exit 1
fi

if [ -d "$WORKDIR_CC" ]; then
    echo "Directory '$WORKDIR_CC' already exists. Remove it or choose another name."
    exit 1
fi

if which CodeChecker >/dev/null 2>&1; then
  echo "CodeChecker found in PATH, remove it!" >&2
  exit 1
fi

# Clone codechecker
git clone https://github.com/Ericsson/codechecker.git "$WORKDIR_CC"
git -C "$WORKDIR_CC" checkout v6.28.2

(
    cd $WORKDIR_CC
    BUILD_UI_DIST=0 make standalone_package
    cd build
    tar -czf ../codechecker_standalone.tar.gz CodeChecker/
)


# Clone zlib
git clone --recurse https://github.com/madler/zlib.git "$WORKDIR"
git -C "$WORKDIR" checkout 5a82f71ed1dfc0bec044d9702463dbdf84ea3b71

# Set Bazel version
echo "8.5.1" > "$WORKDIR/.bazelversion"

# Enable bzlmod
echo "common --enable_bzlmod" > "$WORKDIR/.bazelrc"

# Empty WORKSPACE (required for compat)
touch "$WORKDIR/WORKSPACE"

# Disable requirement for codechecker to be in PATH
cat <<'PATCH' > "$WORKDIR/disable_require_codechecker_in_path.patch"
diff --git a/src/tools.bzl b/src/tools.bzl
index 99a2ea6..16eace1 100644
--- a/src/tools.bzl
+++ b/src/tools.bzl
@@ -20,7 +20,8 @@ Provide tools used by more rulesets.
 def _codechecker_local_repository_impl(repository_ctx):
     codechecker_bin_path = repository_ctx.which("CodeChecker")
     if not codechecker_bin_path:
-        fail("ERROR! CodeChecker is not detected")
+        pass
+        #fail("ERROR! CodeChecker is not detected")
     clang_bin_path = repository_ctx.which("clang")
     if not clang_bin_path:
         fail("ERROR! Clang is not detected")
@@ -38,6 +39,6 @@ def _codechecker_local_repository_impl(repository_ctx):
         executable = False,
     )
 
-    repository_ctx.symlink(codechecker_bin_path, "codechecker_bin")
+    #repository_ctx.symlink(codechecker_bin_path, "codechecker_bin")
     repository_ctx.symlink(clang_bin_path, "clang_bin")
     repository_ctx.symlink(clang_tidy_bin_path, "clang_tidy_bin")
PATCH

# Append rules_codechecker and CodeChecker http_archive to MODULE.bazel
cat <<'MOF_MODULE' >> "$WORKDIR/MODULE.bazel"

git_override(
    module_name = "rules_codechecker",
    remote = "https://github.com/Ericsson/rules_codechecker.git",
    commit = "bed72c8b2efd0cc5ac3ab37840a15656ffa200df",
    # FIX:
    # remote = "https://github.com/furtib/rules_codechecker.git",
    # commit = "0e9307eb5070e170eb1638eeb14014574b5cc4da",
    patches = ["//:disable_require_codechecker_in_path.patch"],
    patch_strip = 1,
)
bazel_dep(name = "rules_codechecker")
bazel_dep(name = "rules_pkg", version = "1.0.1")

http_archive = use_repo_rule(
    "@bazel_tools//tools/build_defs/repo:http.bzl",
    "http_archive",
)
http_archive(
    name = "codechecker",
MOF_MODULE

printf '    urls = ["file://%s/codechecker_standalone.tar.gz"],\n' "$WORKDIR_CC" >> "$WORKDIR/MODULE.bazel"

cat << 'EOF_MODULE' >> "$WORKDIR/MODULE.bazel"
    strip_prefix = "CodeChecker",
    build_file_content = """
filegroup(
    name = "codechecker_files",
    srcs = glob(
        ["**"],
        exclude = ["BUILD.bazel"],
    ),
)

sh_binary(
    name = "codechecker_entry",
    srcs = [":cc_entry_script"],
    data = [":codechecker_files"],
)

genrule(
    name = "cc_entry_script",
    outs = ["cc_entry.sh"],
    cmd = '''
    cat > "$@" <<'EOF'
#!/usr/bin/env bash
set -euo pipefail

if [[ -x "external/+_repo_rules+codechecker/bin/CodeChecker" ]]; then
    exec "external/+_repo_rules+codechecker/bin/CodeChecker" "$$@"
fi

if [[ -x "../+_repo_rules+codechecker/bin/CodeChecker" ]]; then
    exec "../+_repo_rules+codechecker/bin/CodeChecker" "$$@"
fi

script_dir="$$(cd "$$(dirname "$${BASH_SOURCE[0]}")" && pwd)"
if [[ -x "$${script_dir}/bin/CodeChecker" ]]; then
    exec "$${script_dir}/bin/CodeChecker" "$$@"
fi

echo "ERROR: Cannot find CodeChecker binary" >&2
echo "  CWD: $$(pwd)" >&2
echo "  script_dir: $${script_dir}" >&2
exit 1
EOF

chmod +x "$@"
'''
)

""",
)

register_toolchains("//:codechecker_custom_toolchain")
EOF_MODULE

# Append codechecker targets and custom toolchain to BUILD.bazel
cat <<'EOF' >> "$WORKDIR/BUILD.bazel"
#-------------------------------------------------------

# codechecker rules
load(
    "@rules_codechecker//src:codechecker.bzl",
    "codechecker_test",
)

codechecker_test(
    name = "codechecker_test",
    targets = [
        ":z",
    ],
)

codechecker_test(
    name = "per_file_test",
    targets = [
        ":z",
    ],
    per_file = True,
)

#-------------------------------------------------------
load(
    "@rules_codechecker//src:codechecker_toolchain.bzl",
    "codechecker_toolchain",
)

codechecker_toolchain(
    name = "codechecker_custom",
    codechecker = "@codechecker//:codechecker_entry",
)

toolchain(
    name = "codechecker_custom_toolchain",
    toolchain = ":codechecker_custom",
    toolchain_type = "@rules_codechecker//src:toolchain_type",
)
EOF
```

**Expected behaviour**
We expect a successful analysis of zlib, so both codechecker targets to fail.

**Additional context**
This behaviour can also be seen when using pip-installed codechecker (#293), but the issue is not so apparent there.

furtib:
Why:
We currently only provide the single entry-script/binary to tools defined with our toolchain. (In CodeCheckers' case, this means only the entry script is available in the sandbox, not the entire installation of CodeChecker.)
This can cause trouble when users are trying to use the toolchain with their custom Bazel-provided CodeChecker versions.

What:
- Have the toolchain provide the necessary run_files for each target it was defined for. (So in the case of CodeChecker, every file of the project.)

Addresses:
Fixes: #292

Test: #277

### DIFF
--- a/src/codechecker.bzl
+++ b/src/codechecker.bzl
@@ -118,15 +118,13 @@ def _codechecker_impl(ctx):
     ctx.actions.run(
         inputs = depset(
             [
-                info.codechecker,
-                info.clangsa,
-                info.clang_tidy,
                 ctx.outputs.codechecker_script,
                 ctx.outputs.codechecker_commands,
                 ctx.outputs.codechecker_skipfile,
                 config_file,
             ] + source_files,
         ),
+        tools = [info.runfiles],
         outputs = [
             codechecker_files,
             ctx.outputs.codechecker_log,
@@ -250,10 +248,7 @@ def _codechecker_test_impl(ctx):
     # Return test script and all required files
     run_files = default_runfiles + [
         ctx.outputs.codechecker_test_script,
-        info.codechecker,
-        info.clang_tidy,
-        info.clangsa,
-    ]
+    ] + info.runfiles.to_list()
     return [
         DefaultInfo(
             files = depset(all_files),
--- a/src/codechecker_toolchain.bzl
+++ b/src/codechecker_toolchain.bzl
@@ -8,15 +8,34 @@ CodeCheckerInfo = provider(
         "clang_tidy": "clang-tidy executable",
         "clangsa": "Clang executable",
         "codechecker": "CodeChecker executable",
+        "runfiles": "Depset of files needed to run the tools: the three executables " +
+                    "plus their transitive data_runfiles. Pass to `tools` in " +
+                    "ctx.actions.run and include in test runfiles.",
     },
 )
 
 def _codechecker_toolchain_impl(ctx):
+    runfiles = depset(
+        direct = [
+            ctx.executable.codechecker,
+            ctx.executable.clangsa,
+            ctx.executable.clang_tidy,
+        ],
+        transitive = [
+            # We also collect files necessary for these programs to run.
+            # Those files should be declared with `data = [...]`
+            # in the executable's target.
+            ctx.attr.codechecker[DefaultInfo].data_runfiles.files,
+            ctx.attr.clangsa[DefaultInfo].data_runfiles.files,
+            ctx.attr.clang_tidy[DefaultInfo].data_runfiles.files,
+        ],
+    )
     toolchain_info = platform_common.ToolchainInfo(
         codecheckerinfo = CodeCheckerInfo(
             codechecker = ctx.executable.codechecker,
             clang_tidy = ctx.executable.clang_tidy,
             clangsa = ctx.executable.clangsa,
+            runfiles = runfiles,
         ),
     )
     return [toolchain_info]
--- a/src/per_file.bzl
+++ b/src/per_file.bzl
@@ -69,9 +69,6 @@ def _run_code_checker(
             compile_commands_json,
             config_file,
             config,
-            info.codechecker,
-            info.clangsa,
-            info.clang_tidy,
         ] + sources_and_headers
     else:
         # NOTE: we collect only headers, so CTU may not work!
@@ -81,9 +78,6 @@ def _run_code_checker(
             config_file,
             src,
             config,
-            info.codechecker,
-            info.clangsa,
-            info.clang_tidy,
         ], transitive = [headers])
 
     outputs = [
@@ -104,6 +98,7 @@ def _run_code_checker(
         inputs = inputs,
         outputs = outputs,
         executable = ctx.outputs.per_file_script,
+        tools = [info.runfiles],
         arguments = [
             info.codechecker.path,
             data_dir,
@@ -177,6 +172,7 @@ def _create_wrapper_script(ctx, options, compile_commands_json, config_file):
     )
 
 def _per_file_impl(ctx):
+    info = ctx.toolchains["//src:toolchain_type"].codecheckerinfo
     compile_commands = None
     for output in compile_commands_impl(ctx):
         if type(output) == "DefaultInfo":
@@ -235,7 +231,9 @@ def _per_file_impl(ctx):
     files = depset(
         direct = all_files,
     )
-    run_files = [ctx.outputs.test_script, info.codechecker] + all_files
+    run_files = [
+        ctx.outputs.test_script,
+    ] + info.runfiles.to_list() + all_files
     return [
         DefaultInfo(
             files = files,
--- a/test/unit/toolchain/toolchain_runfiles_test.bzl
+++ b/test/unit/toolchain/toolchain_runfiles_test.bzl
@@ -40,15 +40,9 @@ def _subject_impl(ctx):
     tc_info = ctx.attr.toolchain[platform_common.ToolchainInfo]
     info = tc_info.codecheckerinfo
 
-    # TODO: remove hasattr guard once the fix is applied.
-    if hasattr(info, "runfiles"):
-        basenames = sorted([f.basename for f in info.runfiles.to_list()])
-        has_runfiles = True
-    else:
-        basenames = []
-        has_runfiles = False
+    basenames = sorted([f.basename for f in info.runfiles.to_list()])
 
-    return [_RunfilesInfo(basenames = basenames, has_runfiles = has_runfiles)]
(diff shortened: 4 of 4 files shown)


src/codechecker_toolchain.bzl
```diff
@@ -8,15 +8,34 @@ CodeCheckerInfo = provider(
...
+            ctx.executable.clang_tidy,
+        ],
+        transitive = [
+            # We also collect files necessary for these programs to run.
+            # Those files should be declared with `data = [...]`
+            # in the executable's target.
+            ctx.attr.codechecker[DefaultInfo].data_runfiles.files,
+            ctx.attr.clangsa[DefaultInfo].data_runfiles.files,
+            ctx.attr.clang_tidy[DefaultInfo].data_runfiles.files,
+        ],
```
nettle, line 31: I just wonder what dependencies we see here?
Can/should we add a test for this?

src/codechecker.bzl
```diff
@@ -118,15 +118,13 @@ def _codechecker_impl(ctx):
...
-                info.codechecker,
-                info.clangsa,
-                info.clang_tidy,
                 ctx.outputs.codechecker_script,
                 ctx.outputs.codechecker_commands,
                 ctx.outputs.codechecker_skipfile,
                 config_file,
             ] + source_files,
         ),
+        tools = [info.runfiles],
```
nettle, line 127: Looks much nicer!

src/codechecker_toolchain.bzl
furtib: CodeChecker is not a single executable, so we mainly expect its files to be here.
I will try to mock something up using skylib.

nettle: My point is - I think it has no sense to add/change something that does not work.
My question is - if this works what exact dependencies we actually missing?

furtib: In the current state of the toolchain, it only captures the entry script of a standalone codechecker tar.gz (`CodeChecker/bin/CodeChecker`). Any other file (e.g., `CodeChecker/bin/subcmd/CodeChecker-main` which is called by the entry script) is not captured by the toolchain.

Szelethus: >  I think it has no sense to add/change something that does not work.

What do you mean specifically?

Szelethus (commented):
I understand that we are following standard Bazel idioms here, but they might not be that obvious, especially how we expect the `data` field of the toolchain to be populated. Lets document how to add a custom toolchain in README.

I know that we will need to split README up eventually -- but until thats done, throw it there.

Szelethus (changes requested):
I know this is a cheap review, but please create an issue. You can likely reproduce with CodeChecker's [standalone_package target](https://github.com/Ericsson/codechecker#linux-build-from-source).

src/codechecker_toolchain.bzl
nettle: So, do you see `CodeChecker/bin/subcmd/CodeChecker-main` in transitive dependencies? To be honest I have some doubts that transitive works here, it would be really interesting to see...

furtib: I have created a reproducible case for this bug in #292.

furtib:
I added the test from #277.
I have undone the modifications made in 905ae5b328531e294e10a58ba92cd71fdc93f6ba to enable the test.

Szelethus (approved):
Great patch. Especially the testing part of it. Well done, Tibi!

PR #275, opened 2026-08-10
Apply buildifier lint suggestions

furtib:
Why:
Following buildifier's lint suggestions greatly improves Bazel version compatibility.

What:
- Added load statements where we use elements from `rules_cc`. (Ran `buildifier -lint=fix -r .`).
- Did the same changes for `rules_python`. (This means rules_python is again a dependency of the project.)

Addresses:
none


Szelethus (approved):
LGTM

PR #277, opened 2026-08-12
Add skylib analysis test for runfiles for the toolchain

furtib:
Why:
We want a test case for #274.

What:
- Added a skylib analysis test that checks whether runfile dependencies of toolchain binaries are present.

Addresses:
none

### DIFF
--- a/test/unit/toolchain/helper_data.txt
+++ b/test/unit/toolchain/helper_data.txt
@@ -1,1 +1,4 @@
-This file simulates a runtime data dependency of a tool in the toolchain.
+In the BUILD file, this text file is declared as a runfile dependency to the (mock) codechecker toolchain
+through the data field of sh_binary target.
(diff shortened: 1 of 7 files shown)


test/unit/toolchain/helper_data.txt
```diff
@@ -0,0 +1 @@
+This file simulates a runtime data dependency of a tool in the toolchain.
```
Szelethus: Considering the content of this txt file is immaterial, we could expand this a little.

In the BUILD file, this text file is declared as a runfile dependency to the (mock) codechecker toolchain through the `data` field of `sh_binary` target. We expect this text file (and any other, similarly declared resources) to be available through the toolchain.

Szelethus (approved):
LGTM! This was a very clever patch, pleasure to read.

PR #280, opened 2026-08-12
Set use_default_shell_env = False on per_file rule

furtib:
Why:
We want our actions to run without `use_default_shell_env = True`.

What:
- Update the comment on the workaround.
- Set `use_default_shell_env` to false on per_file rule.

Note:
 Fails on Jenkins.

Addresses:
none

### DIFF
--- a/src/codechecker_script.py
+++ b/src/codechecker_script.py
@@ -251,8 +251,9 @@ def analyze(cfg):
         if env_list:
             codechecker_env = dict(item.split("=", 1) for item in env_list)
             env.update(codechecker_env)
+    # Note: This is a workaround, CodeChecker requires the PATH to be set
     if "PATH" not in env:
-        env["PATH"] = "/bin"  # NOTE: this is workaround for CodeChecker 6.24.4
+        env["PATH"] = "/bin"
     env["CC_ANALYZER_BIN"] = generate_analyzer_executables(cfg)
     logging.debug("env: %s", str(env))
 
--- a/src/per_file.bzl
+++ b/src/per_file.bzl
@@ -97,6 +97,7 @@ def _run_code_checker(
     # Action to run CodeChecker for a file
     # env_vars are unused for now, since
     # use_default_shell_env and env are incompatible
+    # TODO: use env for environment variables, instead of passing it as argument
     ctx.actions.run(
         inputs = inputs,
         outputs = outputs,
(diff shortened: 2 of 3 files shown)


src/codechecker_script.py
```diff
@@ -194,7 +194,8 @@ def analyze():
             codechecker_env = dict(item.split("=", 1) for item in env_list)
             env.update(codechecker_env)
     if "PATH" not in env:
-        env["PATH"] = "/bin"  # NOTE: this is workaround for CodeChecker 6.24.4
+        # NOTE: this is workaround for CodeChecker
+        env["PATH"] = "/bin:/usr/bin:/usr/local/bin"
```
Szelethus: According to [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html), in this list, `/bin` is the most fundamental, general directory for binaries, `/usr/bin` comes next and `/usr/local/bin` is the last. `**/sbin` directories are usually, but not always (most notably, [alpine linux](https://www.alpinelinux.org/posts/2025-10-01-usr-merge.html), which is quite frequently used in containers) symlinks to their `**/bin` counterparts.

So, for one, the order of these directories should be reversed from msot special to most general. For two, why shouldn't we include the `sbin` variants if we've already come this far?

furtib: I reordered them, but I don't agree with adding sbin to the PATH.
According to [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s16.html), that folder is for administrator tools. Under no circumstances should we need administrator tools inside a Bazel sandbox. (Won't have permission to run anything from there.)

Szelethus (approved):
LGTM

src/codechecker_script.py
```diff
@@ -194,7 +194,8 @@ def analyze():
             codechecker_env = dict(item.split("=", 1) for item in env_list)
             env.update(codechecker_env)
     if "PATH" not in env:
-        env["PATH"] = "/bin"  # NOTE: this is workaround for CodeChecker 6.24.4
+        # NOTE: this is workaround for CodeChecker
+        env["PATH"] = "/usr/local/bin:/usr/bin:/bin"
```
nettle: This does not look correct to me.
From Bazel perspective we must not do that.
So, why do we?

nettle (changes requested):

src/codechecker_script.py
furtib: ~~After further experimentation, it seems like CodeChecker never actually uses any tool from PATH; it just needs it to be set.~~

I need time for further testing.

furtib:
I have experimented extensively as to why we need this change; there is a more hermetic solution in #308.
Until we can agree on that solution, I propose this change so that we do not use the default shell environment, which is very much an anti-pattern.

nettle (approved):

Issue #44, opened 2025-08-05
`codechecker_suite()` target doesn't work in our tests on ubuntu 22.04


Szelethus:
**Describe the bug**
If I uncomment the `platforms` parameter for target `codechecker_pass_multi` in `test/BUILD`, the target immediately fails with the following error message:

`While resolving toolchains for target //test:codechecker_config_json: Target @platforms//os:linux was referenced as a platform, but does not provide PlatformInfo`

**Versions**
* `CodeChecker version`
```
---------------------------------------------------------------
Kind                 | Version                                 
---------------------------------------------------------------
Base package version | 6.26.0                                  
Package build date   | 2025-06-10T18:03                        
Git commit ID (hash) | 0bbf351e870c7056cd9cc0dc44a795c0fe805dfa
Git tag information  | 6.26                                    
---------------------------------------------------------------

[INFO 2025-08-05 16:23] - CodeChecker web version:
------------------------------------------------------------------------------
Kind                                | Version                                 
------------------------------------------------------------------------------
Base package version                | 6.26.0                                  
Package build date                  | 2025-06-10T18:03                        
Git commit ID (hash)                | 0bbf351e870c7056cd9cc0dc44a795c0fe805dfa
Git tag information                 | 6.26                                    
Server supported Thrift API version | 6.65                                    
Client Thrift API version           | 6.65                                    
------------------------------------------------------------------------------
```
* `bazel --version`: bazel 6.5.0
* codechecker_bazel commit: e5fe33bbfa2ef7e4de18e3a5ca2325dbf368c790 (main branch as of submitting)
* Platform: Ubuntu 22.04

**To Reproduce**
Steps to reproduce the behaviour:
1. Uncomment the platforms field in `test/BUILD`. Here, I only left in the linux platform, but the same issue occurs when I leave in ios and android:
```starlark
codechecker_suite(
    name = "codechecker_pass_multi",
    config = "codechecker_config_json",
    platforms = [
        "@platforms//os:linux",
    ],
    targets = [
        "test_pass",
    ],
)
```
2. Invoke bazel: `bazel test codechecker_pass_multi`

**Expected behaviour**
* The test passes.
* Have a functioning test for this in the CI.
* Have documentation for this rule in README.

**Additional context**
This might stem from my misunderstanding of the rule -- if this is the case, I (we) should try to understand its intent and usage in this open-source sphere alongside industrial applications.

nettle:
[...]

PR #281, opened 2026-08-12
Update Public API

nettle:
Why:
Still we have to reach into the implementation (//src:) for the toolchain:
@rules_codechecker//src:toolchain_type when declaring own toolchain,
and //src:codechecker_local_toolchain when registering ours.
get_platform_alias() was not reachable from defs.bzl either,
so the public API was incomplete.

What:
- Define toolchain_type and the default toolchain in the root package
- Rename codechecker_local to default_tools and
  codechecker_local_toolchain to default_toolchain
- Refer to //:toolchain_type in the rules that resolve the toolchain
- Register //:default_toolchain in MODULE.bazel
- Export get_platform_alias() from defs.bzl

Addresses:
#288

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,3 +1,46 @@
+# Copyright 2026 Ericsson AB
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("//src:codechecker_toolchain.bzl", "codechecker_toolchain")
+
+# The public labels of rules_codechecker live here,
+# the implementation is in //src
+#
+# NOTE: only load from packages visible to our consumers here,
+# this package is loaded by everyone who resolves //:default_toolchain
+
+# Named by convention
+# https://bazel.build/extending/toolchains#writing-rules-toolchains
+toolchain_type(
+    name = "toolchain_type",
+    visibility = ["//visibility:public"],
+)
+
+# Tools found on PATH, provisioned by the default_codechecker_tools extension
+codechecker_toolchain(
+    name = "default_tools",
+    clang_tidy = "@default_codechecker_tools//:clang_tidy",
+    clangsa = "@default_codechecker_tools//:clang",
+    codechecker = "@default_codechecker_tools//:CodeChecker",
+)
+
+toolchain(
+    name = "default_toolchain",
+    toolchain = ":default_tools",
+    toolchain_type = ":toolchain_type",
+)
+
 # Repository root marker, used by the buildifier test to find the workspace
 exports_files(
     ["MODULE.bazel"],
--- a/src/per_file.bzl
+++ b/src/per_file.bzl
@@ -191,8 +191,7 @@ def _per_file_impl(ctx):
     all_files = [compile_commands, config_file]
     _create_wrapper_script(ctx, options, compile_commands, config_file)
 
-    # TODO: Consider using aliases so we don't have to type //src: everywhere.
-    info = ctx.toolchains["//src:toolchain_type"].codecheckerinfo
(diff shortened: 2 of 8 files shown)


src/per_file.bzl
```diff
@@ -192,7 +192,7 @@ def _per_file_impl(ctx):
     _create_wrapper_script(ctx, options, compile_commands, config_file)
 
     # TODO: Consider using aliases so we don't have to type //src: everywhere.
```
Szelethus: There is no need to preserve this comment if this PR lands.

test/buildifier/BUILD
```diff
@@ -0,0 +1,39 @@
+# Copyright 2026 Ericsson AB
```
furtib, line 1: Moving Buildifier into testing is outside the scope of a Public API patch.

BUILD
```diff
@@ -1,15 +1,47 @@
...
+# Named by convention
+# https://bazel.build/extending/toolchains#writing-rules-toolchains
+toolchain_type(
+    name = "toolchain_type",
+    visibility = ["//visibility:public"],
+)
+
+# Tools found on PATH, provisioned by the default_codechecker_tools extension
+codechecker_toolchain(
+    name = "default_tools",
```
furtib, line 32: This name does not communicate that this toolchain cannot be used with remote executors!

This toolchain is specific to a user's local environment.

We could rename it to something like this.
```suggestion
    name = "default_local_tools",
```

Szelethus (changes requested):
Can we [create an issue first](https://github.com/Ericsson/rules_codechecker/blob/main/.github/pull_request_template.md?plain=1#L8) as per the PR requirements? It would help us get a clearer picture for the motivation.

> * Export get_platform_alias() from defs.bzl

That, alongside `codechecker_suite` is poorly documented (related: #44). Also, is this something that should be a public feature of this specific ruleset? Can you expand on the motivation?

> * Move buildifier to test/buildifier

Does this change have to land within this PR? Sounds like a separate addition worhty of a separate discussion.

furtib (changes requested):
The toolchain name should reflect the fact that it is a local-only toolchain.

I don't have strong opinions on the placement of the `toolchain_type`, but we should move it as little as possible. Looking at similar rules, their toolchains are usually like:
- rules_python: [`@rules_python//python:toolchain_type`](https://github.com/bazel-contrib/rules_python/blob/36c342ab1f36a4cda09cd7b1baf8f38c4f1ba6d6/python/BUILD.bazel#L132-L135)
- rules_go: [`@rules_go//go:toolchain_type`](https://github.com/bazel-contrib/rules_go/blob/9fb5d2b23e9a008d4d4d96d7fd210624e98dedd0/go/BUILD.bazel#L40-L43)
- rules_cc: [`@rules_cc//cc:toolchain_type`](https://github.com/bazelbuild/rules_cc/blob/87efbadcecb511f7efa9a063d7a2efb50892e60f/cc/BUILD#L47-L50)
- rules_rust: [`@rules_rust//rust:toolchain_type`](https://github.com/bazelbuild/rules_rust/blob/bb4e9e8cf14511e7b9cf7bda3d1c1f277d076e24/rust/BUILD.bazel#L10-L12)

It is my understanding that these rules use the _rulename_ directory as src, so we do not have to follow their pattern.

Remember that this is an API breaking change.

BUILD
nettle: Well... "local" is actually kinda misleading and wrong here.
Local to what? Besides, should this be default forever?
Default tools are being taken from the environment currently, which is totally wrong!
Later default_tools should became real default, not local.

src/per_file.bzl
nettle: Good point!

test/buildifier/BUILD
nettle: Agree, I've created a separate PR: #289

BUILD
furtib: Local to the user's local/current computer.
If the behaviour of taking tools from the environment isn't something we should communicate to users, or not here, then this name is fine.

Of course, as we will be able to get these dependencies from within Bazel, this toolchain should change.

Why not rename it straight to default? We can change the target binaries as we obtain them in the future. Less API change.

furtib (commented):
Okay, so we decided not to follow the similar projects' *toolname* folder design.

In this case, this PR looks good to me.

Please rebase to take out buildifier related changes, and remove the comment pointed out by @Szelethus!

furtib (approved):
LGTM!

Szelethus (approved):
LGTM

PR #282, opened 2026-08-13
Fix minor issues with toolchain

furtib:
Why:
We want the toolchain to be correct.

What:
- Fixed the wrong default value for CodeChecker.
- Removed `allow_single_file = True` from each target. This attribute prevents users from using `sh_binary` or similar targets to provide their tools. (This is why I needed a mock rule for my test in #277.) 

Addresses:
none


Szelethus (approved):
LGTM

PR #283, opened 2026-08-13
Remove dead code

furtib:
Why:
We have abandoned the two different Python toolchain approaches. Therefore, we do not need this helper function to determine which one we should use.

What:
- Removed `python_toolchain_type` function from `common.bzl`

Addresses:
none


Szelethus (approved):
LGTM

PR #287, opened 2026-08-13
ci: Add github action for tagging every merged PR

furtib:
Why:
We want to tag every commit made to main for easier integration and testing for end users.

What:
- Added a GitHub Action that creates a tag for every merged PR.
	Use the date and time for the tag name.

Addresses:
#285

### DIFF
--- a/.github/workflows/pr-tag.yaml
+++ b/.github/workflows/pr-tag.yaml
@@ -16,9 +16,9 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
       # Optionally create releases too.
-      #- name: Create a GitHub release
-      #  uses: ncipollo/release-action@v1
-      #  with:
-      #    tag: ${{ steps.tag_version.outputs.new_tag }}
-      #    name: Release ${{ steps.tag_version.outputs.new_tag }}
-      #    body: ${{ steps.tag_version.outputs.changelog }}
+      - name: Create a GitHub release
+        uses: ncipollo/release-action@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-tag.yaml
```diff
@@ -0,0 +1,24 @@
+name: Create Tag on Merge
+
+on:
+    pull_request:
+        types: [closed]
+
+jobs:
+    tag:
+        if: github.event.pull_request.merged == true
+        runs-on: ubuntu-latest
```
Szelethus: Do we even need to specify this?

Szelethus (commented):
Consider using an already existing solution like https://github.com/mathieudutour/github-tag-action or https://github.com/anothrNick/github-tag-action.

.github/workflows/pr-tag.yaml
```diff
@@ -0,0 +1,24 @@
...
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                  ref: main
+
+            - name: Create and push tag
+              run: |
+                  TAG_NAME="v$(date +%Y.%m.%d.%H%M%S)"
```
nettle: This is not good idea I think...
Tag as a version number is good, tag as date is not much better than commit hash.

nettle (dismissed):

Szelethus (approved):
LGTM. This is a rather invasive change for our workflow -- if it doesn't work out, we should revert it. But for now, I'm in favour of giving this a shot.

PR #289, opened 2026-08-13
Move buildifier to test/

nettle:
Why:
buildifier_prebuilt is a dev dependency, therefore it is not visible to
the users while the root package has to stay loadable by them.
A lint test also belongs the tests rather than in the root package.

What:
- Move buildifier_test and its patches to test/buildifier
- Export MODULE.bazel to the linter package, it is the workspace marker
- Point the buildifier_prebuilt patches to the new location
- Run //test/buildifier in the lint step of the workflow


furtib (approved):
LGTM!

PR #293, opened 2026-08-17
Demo pip codechecker

furtib:
This PR is a bit premature.

Why:
We want a hermetic toolchain.

What:
- Added a way to obtain codechecker through pip.
- For some reason, we need to set the `CC_BIN_DIR` environment variable to point to the codechecker's config directory. (This part needs some work... as you can see on CI)

Notes:
Depends on: #274

Addresses:
none?


requirements_lock.txt
```diff
@@ -0,0 +1,47 @@
+alembic==1.19.1
```
nettle, line 1: OMG, is all of this hm.. stuff really needed for CodeChecker?

MODULE.bazel
```diff
@@ -39,6 +39,22 @@ codechecker_extension = use_extension(
...
+    python_version = "3.12",
+)
+
+pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
+pip.parse(
+    hub_name = "pip",
+    python_version = "3.12",
+    requirements_lock = "//:requirements_lock.txt",
+)
+use_repo(pip, "pip")
```
nettle, line 60: Yes, this is how it should be done

nettle (commented):
I was thinking about something like this!

furtib:
Okay, now it passes in CI.
I needed to add a new element to the toolchain: `codechecker_files_to_run`.
I cannot integrate this into the run_files depset because its content is a different type, even though its a very similar concept.

PR #294, opened 2026-08-17
feat: Add option to override toolchain on a per target basis

### LINKED ISSUE #278: Per-target toolchain selection argument
### Describe the task

Users want to be able to specify a specific toolchain to use at a target-specific level.

### Scope

- Add a new argument that can specify the toolchain to use for a specific `codechecker_test` action.
- Document changes.

### Definition of Done

- It is possible for users to define a toolchain to use on a per-target basis.
- Changes are documented.

furtib:
Why:
Users want the ability to use different versions of CodeChecker (or analyzers) on a per-target basis.

What:
- Added a `codechecker_toolchain` parameter to `codechecker_test`, through which users can override Bazel's toolchain resolution. (Preserves previous behaviour if not specified.)

Addresses:
Fixes: #278

### DIFF
--- a/src/codechecker.bzl
+++ b/src/codechecker.bzl
@@ -23,6 +23,7 @@ load(
 )
 load(
     "common.bzl",
+    "resolve_toolchain_info",
     "version_specific_attributes",
 )
 load(
@@ -92,7 +93,7 @@ def _codechecker_impl(ctx):
 
     config_file, codechecker_env = get_config_file(ctx)
 
-    info = ctx.toolchains["//:toolchain_type"].codecheckerinfo
+    info = resolve_toolchain_info(ctx)
 
     codechecker_files = ctx.actions.declare_directory(ctx.label.name + "/codechecker-files")
 
@@ -171,6 +172,12 @@ codechecker = rule(
             default = [],
             doc = "List of analyze command arguments, e.g.; --ctu.",
         ),
+        "codechecker_toolchain": attr.label(
+            default = None,
+            doc = "Optional codechecker_toolchain() target. " +
+                  "When set, tools from this target are used instead of " +
+                  "Bazel's toolchain resolution.",
+        ),
         "config": attr.label(
             default = None,
             doc = "CodeChecker configuration",
@@ -227,7 +234,7 @@ def _codechecker_test_impl(ctx):
     if not codechecker_files:
         fail("Execution results required for codechecker test are not available")
 
-    info = ctx.toolchains["//:toolchain_type"].codecheckerinfo
+    info = resolve_toolchain_info(ctx)
 
     # Create test script
     codechecker_test_script = ctx.actions.declare_file(ctx.label.name + "/codechecker_test_script")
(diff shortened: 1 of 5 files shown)


src/codechecker.bzl
```diff
@@ -92,7 +93,7 @@ def _codechecker_impl(ctx):
 
     config_file, codechecker_env = get_config_file(ctx)
 
-    info = ctx.toolchains["//src:toolchain_type"].codecheckerinfo
+    info = resolve_toolchain_info(ctx)
```
nettle: Kind of doubt that we need a function here...

```py
if ctx.attr.toolchain:
    info = ctx.attr.toolchain[platform_common.ToolchainInfo].codecheckerinfo
else:
    info = ctx.toolchains["//:toolchain_type"].codecheckerinfo
```

src/codechecker.bzl
```diff
@@ -174,6 +175,12 @@ codechecker = rule(
             default = [],
             doc = "List of analyze command arguments, e.g.; --ctu.",
         ),
+        "codechecker_toolchain": attr.label(
```
nettle: Maybe just "toolchain"?

src/codechecker.bzl
```diff
@@ -229,9 +236,7 @@ def _codechecker_test_impl(ctx):
     if not codechecker_files:
         fail("Execution results required for codechecker test are not available")
 
-    info = ctx.toolchains["//src:toolchain_type"].codecheckerinfo
-
-    # Create test script from template
```
nettle, line 232: I guess removed by mistake?

furtib: Yes, thanks for pointing it out!

src/codechecker.bzl
furtib: Yeah, that makes much more sense.

src/codechecker.bzl
furtib: I wanted to avoid code duplication as much as possible (we have to do this for all 3 rules: `codechecker`, `codechecker_test` and `pre_file_test`), but I guess we do not necessarily need it.

furtib (commented):
Thank you for the review!

src/codechecker.bzl
Szelethus: What is the drawback behind small, tidy functions that are reused? I'd strongly prefer the earlier solution.

nettle: The main drawback of small functions is that excessive breaking up of code increases cognitive overhead and makes it harder to follow the overall logic. The same for splitting functionality into multiple files - that increases navigation complexity, which forces you to jump between files just to trace how a feature works. Besides it makes solution less self-contained and tend to add unnecessary dependencies (tight coupling) which also leads to loosing track of them - we already have it (tools.bzl, common.bzl, codechecker_toolchain.bzl). Ultimately, a balanced and very reasonable approach should be taken when breaking down code into functions and files, and not just DRY.

Szelethus (commented):
LGTM from the code's perspective -- should we not bother with a skylib test?

Szelethus (approved):
LGTM on my end. These skylib tests are quite neat.

PR #295, opened 2026-08-18
Fix toolchain path for per_file test action

furtib:
Why:
The API change branch was created before #266 landed and was not rebased, so one instance of accessing the toolchain was not rewritten.

What:
- Updated the toolchain type location used in the per_file_test rule.

Addresses:
CI failing on main


Szelethus (approved):
LGTM

PR #296, opened 2026-08-18
Add more from compile_commands.bzl to public API

nettle:
Why:
Users write their own rules on top of our compilation database collection,
but compile_commands_aspect, SourceFilesInfo and platforms_transition were
only reachable from src/compile_commands.bzl.

What:
- Add compile_commands_aspect, SourceFilesInfo and platforms_transition to defs.bzl


furtib (approved):
LGTM!

Issue #156, opened 2026-01-20
Split README to simple landing page and documentation


nettle:
### Describe the task

We should split our long README.md into simple landing page and actually documentation.
Reason: README.md should be very simple (What it is, How to use, that's it).
For documentation I would suggest https://github.com/docsifyjs/docsify or similar
 

### Scope

1. Move current README.md to doc/README.md
2. Create small main README.md with minimal information:
   - What it is
   - How to use
3. Optionally add simple documentation rendering engine
   - One of the options: docsify
4. Optionally add documentation publishing to CI
5. Plan next for documentation development

### Definition of Done

- Simple README.md
- Documentation under doc/

PR #297, opened 2026-08-19
Refactor documentation

nettle:
Why:
The documentation was one 500 line page, no structure.
The contribution page mixed reporting issues with the developer setup.

What:
- Split docs/README.md into pages
- Support static rendering with Docsify 5
- Add pages for the public API, troubleshooting and contributing
- Reduce CONTRIBUTING.md to reporting issues and a pointer to the docs
- Update the contents to correspond current status

Addresses:
#156


docs/toolchains.md
```diff
@@ -0,0 +1,68 @@
+Toolchains
+==========
+
+The rules locate the CodeChecker, Clang and clang-tidy executables through a
+Bazel [toolchain](https://bazel.build/extending/toolchains). Modeling the tools
+as a toolchain means the analysis rule (`codechecker_test`) never hardcode binary paths:
```
furtib:
```suggestion
as a toolchain means the analysis rule (`codechecker_test`) never hardcodes binary paths:
```

docs/troubleshooting.md
```diff
@@ -0,0 +1,56 @@
...
+
+```bash
+which clang gcc
+```
+
+
+## Sandboxing
+
+> [!WARNING]
+> When debugging, or reporting a bug, take into consideration which sandbox mode did bazel ran in. Be careful as the default sandbox can change if you ran [Bazel inside Docker](https://bazel.build/versions/6.5.0/docs/sandboxing#:~:text=Both%20the%20linux,run%20%2D%2Dprivileged.).
```
furtib:
```suggestion
> When debugging, or reporting a bug, take into consideration which sandbox mode bazel ran in. Be careful as the default sandbox can change if you ran [Bazel inside Docker](https://bazel.build/versions/6.5.0/docs/sandboxing#:~:text=Both%20the%20linux,run%20%2D%2Dprivileged.).
```

furtib (approved):
Looks great!
I was unable to open the HTML file locally, I needed to start a webserver to be able to view it!
Are we / Should we create a github pages for this html page, so users can view this more easily?

nettle:
> I was unable to open the HTML file locally, I needed to start a webserver to be able to view it!

Hm... that should work as a static page, no webserver is required.
I'll check...

> Are we / Should we create a github pages for this html page, so users can view this more easily?

Yes, exactly, the simplest way: just publish `docs/` as https://ericsson.github.io/rules_codechecker/
(old example: https://nettle.github.io/documentation-template)

nettle:
> > I was unable to open the HTML file locally, I needed to start a webserver to be able to view it!
> 
> Hm... that should work as a static page, no webserver is required. I'll check...

Modern web browsers block loading resources (mostly scripts) when opening local HTML files considering those as potential security threats:
```
Unsafe attempt to load URL file:///.../index.html from frame with URL file://.../index.html.
'file:' URLs are treated as unique security origins.
```
So, yes, unfortunately we cannot see this as local file

docs/toolchains.md
nettle: Done

docs/troubleshooting.md
nettle: Done

Szelethus (changes requested):
Mostly OK.

1. What is the rationale behind renaming README.md to index.md?
2. Do you insist on merging Docsify 5 in the same PR?

nettle:
> 1. What is the rationale behind renaming README.md to index.md?

The main reason is that we could still use README.md for repo level description and not as the entry point for the Documentation. Just adds a bit more flexibility. Do you see what I mean?
Anyway, this is optional and we can change it back to default README.md

> 2. Do you insist on merging Docsify 5 in the same PR?

I guess by Docsify 5 you mean index.html file, right?
Well, it is just a simplest rendered which we may use to publish ready-to-read documentation as GitHub pages.
Sure, we can create a different PR if it has any sense.

I can do a simple demo

Szelethus (approved):
I'm not a fan of the README->index change, but otherwise, this is a great patch. As the benefits of merging this now far outweights the costs of delaying this over that debate (or merging index.html separately), lets land it!

PR #300, opened 2026-08-20
Experiment: FOSS as external modules

nettle:
Experimenting with new approach: FOSS repos as external Bazel modules

Why:
The existing Python/Bash FOSS tests are hard to
understand and extend, they are also very slow
and cannot be integrated into Bazel environment.

What:
Add test/bazel/ with a foss_test() macro
that generates py_test() targets for FOSS projects.

Test:
bazel test //test/bazel/...

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -154,6 +154,11 @@ jobs:
           runuser -u test -- bazel test //test/buildifier --test_output=all
         shell: bash -el {0}
 
+      - name: Run foss tests
+        run: |
+          runuser -u test -- bazel test //test/bazel/... --test_output=all
+        shell: bash -el {0}
+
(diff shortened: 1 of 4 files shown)


furtib:
This solution is much slower (360sec) than the one in #209 (120sec).
I don't fully understand why, but that solution in particular is fast; any other like this one, or #235, runs really slowly.

nettle:
> This solution is much slower (360sec) than the one in #209 (120sec). I don't fully understand why, but that solution in particular is fast; any other like this one, or #235, runs really slowly.

The reason is simple - we download, extract and run bazel build not once (`setUpClass`) but for every test (`setUp`).
Only in #209 this is fixed.

furtib:
Superseded by #209

PR #301, opened 2026-08-24
test: per-target toolchain selection tests

furtib:
Why:
We want test for our new feature.

What:
- Added test for overwriting toolchain on a per target bases.

Addresses:
#278

### DIFF
--- a/test/unit/toolchain/BUILD
+++ b/test/unit/toolchain/BUILD
@@ -52,13 +52,36 @@ sh_binary(
     srcs = ["mock_clang_tidy.sh"],
 )
 
-# --- Mock toolchain ---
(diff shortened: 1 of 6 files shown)


Szelethus (commented):
When you write standard, you mean "monolothic", right? Because this is the first time I see this term for it in the repo.

Also, shouldn't we have a target that has `runfiles_mock_toolchain1`, and another with `runfiles_mock_toolchain2`, and check whether it was correctly passed along to the target?

furtib:
Okay, renamed and added a second test using the second mock toolchain.

Szelethus (approved):
LGTM

PR #303, opened 2026-08-24
fix: Compute compile flags once per target

### LINKED ISSUE #304: Quote includes are not transitively collected from implementation deps dependencies
**Describe the bug**
C/C++ compile flags named `quote_includes` are not collected from implementation_deps dependencies.
This can cause a compilation error during analysis.

**Versions**
* `CodeChecker 6.28.2`
* `bazel 8.5.1`
* ba752b067ea7beaab40916be00f7e1837ad7436e

**To Reproduce**
Run the skylib tests introduced in #284.

**Expected behaviour**
The mock attribute, to not be in the compilation database.

### LINKED ISSUE #305: Defines are not transitively collected from implementation_deps dependencies.
**Describe the bug**
C/C++ compile flags named `defines` are not collected from implementation_deps dependencies.
This can cause a compilation error during analysis.

**Versions**
* `CodeChecker 6.28.2`
* `bazel 8.5.1`
* ba752b067ea7beaab40916be00f7e1837ad7436e

**To Reproduce**
Run the skylib tests introduced in #284.

**Expected behaviour**
The mock attribute, to be in the compilation database.

nettle:
Why:
Dependency flags, copts and built-in include directories
were recomputed for every source file of a target,
and get_compile_flags mixed two concerns:
flags of the given CcInfo target and walking deps of ctx.rule.attr.
That made the handling of implementation_deps hard to follow.

What:
- Split to get_compile_flags() and get_implementation_deps_flags()
- Collect dependency flags once per target and drop duplicates
- Move copts and built-in includes out of the per source loop
- Do not collect local_defines for implementation_deps
- Drop unused target parameter, compile_flags, compile_variables
- Add the once pattern to unit_test() to detect duplications
- Add tests for implementation_deps, flag propagation via deps,
  local_defines and deduplication

Addresses:
Fixes: #304 #305
#307

### DIFF
--- a/src/compile_commands.bzl
+++ b/src/compile_commands.bzl
@@ -80,14 +80,16 @@ SYSTEM_INCLUDE = "-isystem "
 QUOTE_INCLUDE = "-iquote "
 EXTERNAL_INCLUDE = "-isystem "
 
-# Function copied from https://gist.github.com/oquenchil/7e2c2bd761aa1341b458cc25608da50c
-# NOTE: added local_defines
-def get_compile_flags(ctx, dep):
+def get_compile_flags(dep, local_defines = True):
     """ Return a list of compile options
 
+    Original function copied from
+    https://gist.github.com/oquenchil/7e2c2bd761aa1341b458cc25608da50c
+    Also added local_defines and external_includes
+
     Args:
-        ctx: The context variable.
         dep: A target with CcInfo.
+        local_defines: if local_defines should be added (default: True)
     Returns:
       List of compile options.
     """
@@ -97,8 +99,9 @@ def get_compile_flags(ctx, dep):
     for define in compilation_context.defines.to_list():
         options.append("-D'{}'".format(define))
 
-    for define in compilation_context.local_defines.to_list():
-        options.append("-D'{}'".format(define))
+    if local_defines:
+        for define in compilation_context.local_defines.to_list():
+            options.append("-D'{}'".format(define))
 
     for system_include in compilation_context.system_includes.to_list():
         if len(system_include) == 0:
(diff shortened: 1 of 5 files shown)


src/compile_commands.bzl
```diff
@@ -114,42 +113,37 @@ def get_compile_flags(ctx, dep):
...
+            if type(deps) == "list":
+                for dep in deps:
+                    if CcInfo in dep:
+                        compilation_deps.append(dep)
+    deps_flags = []
+    for dep in compilation_deps:
+        for flag in get_compile_flags(dep):
+            if flag not in deps_flags:
+                deps_flags.append(flag)
+    return deps_flags
```
furtib: Bazel already propagates flags for every SOURCE_ATTR type except implementation_deps (that's the point of implementation_deps). We don't need to loop over every attribute, only implementation_deps.
See my changes: https://github.com/Ericsson/rules_codechecker/pull/272/changes#diff-057ef75766b6e979efc7e1e8a4fc41ce228ae8465ee228e8d45c6d66983dcb94R147.

Something like this suggestion should work correctly:
```suggestion
	# implementation_deps intentionally does not propagate its includes
    # into the target's CcInfo.compilation_context, so we must extract
    # them explicitly. This may produce duplication.
    if not hasattr(ctx.rule.attr, "implementation_deps"):
        return []

    impl_deps = getattr(ctx.rule.attr, "implementation_deps")
    if type(impl_deps) != "list":
        return []

    deps_flags = []
    for dep in impl_deps:
        if CcInfo not in dep:
            continue
        for flag in get_compile_flags(dep):
            if flag not in deps_flags:
                deps_flags.append(flag)
    return deps_flags
```

furtib (commented):
If I understand correctly, you also outsourced the flag collection to a single function and call implementations on that too, fixing the flag mismatches.
But you also streamlined the very start of the collection so we only compute the flags once per target.
This means #272 will be superseded by this one.
I have one minor nit about looping over all SOURCE_ATTR labels; otherwise LGTM!

src/compile_commands.bzl
nettle: I think you are right, this "implementation_deps" is a strange thing.
To be honest I still dont understand how it works and what is the difference in dependencies.
I will take deeper look.

furtib (approved):
LGTM!

PR #308, opened 2026-08-25
feat: Create fake path for codechecker

furtib:
Why:
We want the Bazel rule to be as hermetic as possible.

What:
- Added dirname to the toolchain. (This is the only tool we use from PATH. Running codechecker with strace reveals that it accesses openssl and uname too, but they seem to be optional)
- Added a custom PATH containing both the hermetic Python interpreter (even though the script running CodeChecker is a py_binary target, CodeChecker still wants to find Python in PATH) and dirname.

Addresses:
none

Notes:
* We might be able to remove Python from the fake PATH when switching to pip-installed CodeChecker.
Supporting standalone builds of CodeChecker, however, will still require Python in PATH.

### DIFF
--- a/src/codechecker.bzl
+++ b/src/codechecker.bzl
@@ -93,6 +93,8 @@ def _codechecker_impl(ctx):
     config_file, codechecker_env = get_config_file(ctx)
 
     info = ctx.toolchains["//:toolchain_type"].codecheckerinfo
+    py_toolchain = ctx.toolchains["@rules_python//python:toolchain_type"]
+    py_interpreter_dir = py_toolchain.py3_runtime.interpreter.dirname
 
     codechecker_files = ctx.actions.declare_directory(ctx.label.name + "/codechecker-files")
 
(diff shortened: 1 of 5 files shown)


nettle (changes requested):
I think we are just missing evidences here (use-case? tests?).
That's why this change does not look convincing to me so far.
Is it a part bigger change?

furtib:
I'm not sure I understand your concerns.
This is not part of a bigger change.

The main goal of this patch is to move away from using `use_default_shell_env = True` and get rid of the workaround of setting the PATH to `/bin` to use system tools.
The use case is to have *everything* available for codechecker inside the sandbox (just like with clang, and clang-tidy). (In this case, that would be `dirname` and python)

On the note of tests, since I'm only changing the environment codechecker runs in, any test running a codechecker analysis (e.g.: `//test/unit/basic:basic_codechecker_test` or `//test/unit/basic:basic_per_file_test`) should suffice, since codechecker either works or not.

nettle:
> The main goal of this patch is to move away from using `use_default_shell_env = True` and get rid of the workaround of setting the PATH to `/bin` to use system tools. The use case is to have _everything_ available for codechecker inside the sandbox (just like with clang, and clang-tidy). (In this case, that would be `dirname` and python)

OK, it's getting clearer... Thanks for explanation!
So, we get back python toolchain for that :)
But I still dont understand that idea with fake_path... looks kinda clumsy :)

PR #310, opened 2026-08-25
fix: Docsify sidebar for mobile devices

nettle:
Why:
We should hide the sidebar for narrow screens
and show toggle button to bring sidebar back.
On wider screens the sidebar should be always visible.

What:
- Hide the toggle button only above the breakpoint
- Drop the body.close rule, docsify 5 uses .sidebar.show instead


furtib (approved):
LGMT!

PR #311, opened 2026-08-26
refactor: Migrate per-file script to use argparse

furtib:
Why:
Argparse gives us more control over arguments to the Python script. This change would prevent us from having to constantly update the number of expected arguments.

What:
- Refactored per_file script to use argparse

Addresses:
none


Szelethus (approved):
LGTM, good idea to kep it consistent with the monolithic wrapper

PR #312, opened 2026-08-26
test: Migrate plist_res test into a bazel test

furtib:
Why:
We want all of our tests to be integrated into Bazel.

What:
- Migrated the plist resolution test into a bazel py_test.

Addresses:
#210


Szelethus (changes requested):

Szelethus (approved):
LGTM

PR #313, opened 2026-08-26
refactor: Migrate cache tests to bazel test

furtib:
Why:
We want all tests to run with `bazel test //...`.

What:
- Migrated the cache test to a Bazel test.

Addresses:
#210


Szelethus (approved):
I'm not thrilled about moving the source files into the test, but neither am I against it. Its just not consistent with other tests.

The actual BUILD file is great at self-documenting, though its kind of painful how much baggage must files like `caching_check.py` carry.

PR #314, opened 2026-08-26
chore: Remove unused files

furtib:
Why:
We don't want clutter inside the repository.

What:
- Remove unused test.sh
- Remove .ci/mise
- Remove unnecessary `__init__.py` files (Others are necessary for the current pylint setup, and should be removed in the patch that replaces it.)
- Remove requirements.txt (we use micromamba for dependency management)

Addresses:
none


Szelethus (approved):
LGTM

PR #315, opened 2026-08-26
feat: Make per_file rule support severities

### LINKED ISSUE #309: Add severities attribute to per_file
### Describe the task

We should have parity between the monolithic and per_file rules.
The per_file version of the rule currently does not accept severities as attributes and fails even for low-priority issues.

### Scope

- Refactor per_file into a build and a test rule. (CodeChecker should only be called in the build rule, just as in the monolithic rule)
- Add a severities attribute to the pre_file rule. (Handle it the same way as monolithic, without CodeChecker)

### Definition of Done

- Users can define severities for the per_file test target too.

furtib:
Development halted until issues with integration are ironed out. #199 
Why:
We want complete feature parity between the monolithic and the per_file rule.
The per_file rule currently does not support the severities attribute.

The per_file rule should behave exactly the same during parsing, and when checking the severities, any difference between the two should be considered a bug, not a feature.
That's why, instead of rewriting the logic for parsing and checking severities, I have reused the well-tested monolithic implementation.

What:
- Refactored necessary functions out of the codechecker script into a shared common library.
- Created a parse step for the per_file rule, reusing the one used for the monolithic version.
- Created a test step for the per_file rule, reusing the severity-checking function of the monolithic rule.

Addresses:
Fixes: #309
Fixes: #279 

Depends on: #311

### DIFF
--- a/test/unit/caching/BUILD
+++ b/test/unit/caching/BUILD
@@ -33,14 +33,14 @@ caching_test(
 
 caching_test(
     name = "caching_per_file_test",
-    expected_action_count = 1,
+    expected_action_count = 2,
     file_to_modify = "secondary.cc",
     target_name = "per_file_caching",
 )
 
 caching_test(
     name = "caching_per_file_ctu_test",
-    expected_action_count = 2,
+    expected_action_count = 3,
     file_to_modify = "secondary.cc",
     target_name = "per_file_caching_ctu",
 )
--- a/src/BUILD
+++ b/src/BUILD
@@ -12,6 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 load("@rules_python//python:py_binary.bzl", "py_binary")
+load("@rules_python//python:py_library.bzl", "py_library")
+
+py_library(
+    name = "common",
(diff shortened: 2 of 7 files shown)


Szelethus (commented):
So, much of this PR is completely trivial refactoring. I'd prefer to have the NFC parts land separately, because it'd make the review easier. Of substance, the patch looks mostly right, and I agree with the overall direction of reusing more code. Speaking of that, I wish we could resurrect #141...

test/unit/caching/BUILD
```diff
@@ -33,14 +33,14 @@ caching_test(
...
-    expected_action_count = 1,
+    expected_action_count = 2,
     file_to_modify = "secondary.cc",
     target_name = "per_file_caching",
 )
 
 caching_test(
     name = "caching_per_file_ctu_test",
-    expected_action_count = 2,
+    expected_action_count = 3,
```
Szelethus: What are we seeing here?

furtib: The parse action is a separate action, thus, the action count is increased, as expected.

PR #317, opened 2026-08-26
ci: Remove Release prefix from release name

furtib:
Why:
We want the release name to be the same as the tag name.

What:
- Changed the name parameter for the release from `Release <tagname>` to `<tagname>`

Addresses:
none


Szelethus (changes requested):
Sorry, my remark was confusing. I had only issues with the "Release " prefix. It should only be the tag name, like the other releases.

Szelethus (approved):
sweet

PR #318, opened 2026-08-27
test: integrate compile_flags tests into the skylib test

furtib:
Why:
We want to integrate all of our tests into Bazel. The unit test macro is unnecessarily bulky for tests written in `compile_flags`; they can be checked much more easily and quickly using Skylib tests.

What:
- removed compile_flags test
- Added conly and cxxonly flag tests to the Skylib tests

Addresses:
#210


Szelethus (commented):
Why not combine the 4 tests into two? For the `*.c` file, assert that `__CXX__` is absent and `__CONLY__` is present, and likewise for `*.cc`. That would be more expressive as well IMO.

Szelethus (approved):
LGTM

PR #319, opened 2026-08-27
feat: Add platform to per_file rule

furtib:
Why:
We want the per_file rule to be a drop-in replacement for the monolithic rule.
For this, every attribute of the monolithic rule should be supported in the per_file rule too.

What:
- Added a platform attribute to per_file.  

Addresses:
none

### DIFF
--- a/src/per_file.bzl
+++ b/src/per_file.bzl
@@ -20,10 +20,15 @@ for each translation unit.
 load("@rules_cc//cc/common:cc_info.bzl", "CcInfo")
 load("codechecker_config.bzl", "get_config_file")
 load(
+    "common.bzl",
+    "version_specific_attributes",
+)
+load(
     "compile_commands.bzl",
     "SourceFilesInfo",
     "compile_commands_aspect",
     "compile_commands_impl",
+    "platforms_transition",
 )
 
 # buildifier: disable=unused-variable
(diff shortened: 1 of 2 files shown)


src/per_file.bzl
```diff
@@ -19,11 +19,16 @@ for each translation unit.
 
 load("@rules_cc//cc/common:cc_info.bzl", "CcInfo")
 load("codechecker_config.bzl", "get_config_file")
+load(
+    "common.bzl",
+    "version_specific_attributes",
+)
```
Szelethus: Mind that this function is only for Bazel <=6, which we no longer officially support. There is a decision to be made fornew features in between "might work" and outright "wont even budge" on non-supported versions. I think I'd lead towards the latter.

Szelethus (commented):
The PR is mostly trivial, just a question about older bazel versions.

src/per_file.bzl
furtib: I think that the fate of the `version_specific_attributes` flag is out of scope for this patch.
This should be handled in a separate ticket or PR.

Szelethus (approved):
LGTM, though we should eventually shed the overhead of bazel 6. Might be nicer to do it before landing this, but it doesn't matter much. Lets land it.

